PR #2, opened 2015-10-28
DOCS: Update readthedocs settings

mgeier:
We should do this once matplotlib 1.5 is out and available on readthedocs.org.


mgeier:
Since readthedocs is very slow in upgrading pre-installed packages and since requesting a newer version via `requirements.txt` doesn't seem to work, I've changed the strategy. Now `conda` is used, which can successfully install matplotlib 1.5+.

PR #3, opened 2015-11-01
Create a ndarray subclass for grids and vector fields

mgeier:
WARNING: this does not work in all cases!

One example where it doesn't work is the `np.inner()` function.

I'm doing this PR just for documenting the (failed) attempt to create an ndarray subclass.

It looks like we're better of just to use a plain `np.ndarray(..., dtype=object)` ...


mgeier:
It seems to work after all if we allow `ndim == 0` in `__array_finalize__()` ...

PR #6, opened 2016-01-13
New mono dipol point source

hagenw:
I added a dipol point source after the [Matlab implementation from Fiete](https://github.com/sfstoolbox/sfs/blob/master/SFS_monochromatic/greens_function_mono.m#L89).

### DIFF
--- a/sfs/mono/source.py
+++ b/sfs/mono/source.py
@@ -92,7 +92,7 @@ def point_dipole(omega, x0, n0, grid, c=None):
         Position of source.
     n0 : (3,) array_like
         Normal vector (direction) of dipole.
-    grid : triple of numpy.ndarray
+    grid : triple of array_like
         The grid that is used for the sound field calculations.
         See :func:`sfs.util.xyz_grid`.
     c : float, optional
@@ -145,7 +145,7 @@ def point_modal(omega, x0, n0, grid, L, N=None, deltan=0, c=None):
     n0 : (3,) array_like
         Normal vector (direction) of source (only required for
         compatibility).
-    grid : triple of numpy.ndarray
+    grid : triple of array_like
         The grid that is used for the sound field calculations.
         See :func:`sfs.util.xyz_grid`.
     L : (3,) array_like
@@ -216,7 +216,7 @@ def point_modal_velocity(omega, x0, n0, grid, L, N=None, deltan=0, c=None):
     n0 : (3,) array_like
         Normal vector (direction) of source (only required for
         compatibility).
-    grid : triple of numpy.ndarray
+    grid : triple of array_like
         The grid that is used for the sound field calculations.
         See :func:`sfs.util.xyz_grid`.
     L : (3,) array_like
--- a/sfs/mono/source.py
+++ b/sfs/mono/source.py
@@ -26,6 +26,8 @@ from .. import defs
 def point(omega, x0, n0, grid, c=None):
     """Point source.
 
+    Notes
+    -----
     ::
 
                       1  e^(-j w/c |x-x0|)
@@ -283,6 +285,8 @@ def line(omega, x0, n0, grid, c=None):
 
     Note: third component of x0 is ignored.
 
+    Notes
+    -----
     ::
 
                            (2)
@@ -350,6 +354,8 @@ def line_dipole(omega, x0, n0, grid, c=None):
 
     Note: third component of x0 is ignored.
 
+    Notes
+    -----
     ::
 
                            (2)
@@ -371,6 +377,8 @@ def line_dipole(omega, x0, n0, grid, c=None):
 def plane(omega, x0, n0, grid, c=None):
     """Plane wave.
 
+    Notes
+    -----
     ::
 
         G(x, w) = e^(-i w/c n x)
@@ -397,6 +405,8 @@ def plane(omega, x0, n0, grid, c=None):
 def plane_velocity(omega, x0, n0, grid, c=None):
     """Velocity of a plane wave.
 
+    Notes
+    -----
     ::
 
         V(x, w) = 1/(rho c) e^(-i w/c n x) n
--- a/sfs/mono/source.py
+++ b/sfs/mono/source.py
@@ -79,6 +79,58 @@ def point_velocity(omega, x0, n0, grid, c=None):
     return util.XyzComponents([v * o / r for o in offset])
 
 
+def point_dipole(omega, x0, n0, grid, c=None):
+    """Point source with dipole characteristics.
+
+    Parameters
+    ----------
+    omega : float
+        Frequency of source.
+    x0 : (3,) array_like
+        Position of source.
+    n0 : (3,) array_like
+        Normal vector (direction) of dipole.
+    grid : triple of numpy.ndarray
+        The grid that is used for the sound field calculations.
+        See :func:`sfs.util.xyz_grid`.
+    c : float, optional
+        Speed of sound.
+
+    Returns
+    -------
+    numpy.ndarray
+        Sound pressure at positions given by `grid`.
+
+    Notes
+    -----
+    ::
+
+         d                1   / iw       1    \   (x-x0) n0
+        ---- G(x-x0,w) = --- | ----- + ------- | ----------- e^(-i w/c |x-x0|)
+        d ns             4pi  \  c     |x-x0| /   |x-x0|^2
+
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        n0 = 0, 1, 0
+        p = sfs.mono.source.point_dipole(omega, x0, n0, grid)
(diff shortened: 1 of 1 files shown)


sfs/mono/source.py
```diff
@@ -79,6 +79,54 @@ def point_velocity(omega, x0, n0, grid, c=None):
     return util.XyzComponents([v * o / r for o in offset])
 
 
+def point_dipol(omega, x0, n0, grid, c=None):
```
mgeier: dipole

sfs/mono/source.py
```diff
@@ -79,6 +79,54 @@ def point_velocity(omega, x0, n0, grid, c=None):
...
+
+    Parameters
+    ----------
+    omega : float
+        Frequency of source.
+    x0 : (3,) array_like
+        Position of source.
+    n0 : (3,) array_like
+        Normal vector (direction) of dipole.
+    grid : triple of numpy.ndarray
```
mgeier: This should actually be `triple of array_like`.

This should also be changed in `point_modal*()` ...

sfs/mono/source.py
```diff
@@ -79,6 +79,54 @@ def point_velocity(omega, x0, n0, grid, c=None):
...
+        The grid that is used for the sound field calculations.
+        See :func:`sfs.util.xyz_grid`.
+    c : float, optional
+        Speed of sound.
+
+    Returns
+    -------
+    numpy.ndarray
+        Sound pressure at positions given by `grid`.
+    ::
```
mgeier: If the equation is supposed to be part of the return value docs, it has to be indented.

If not (which I think I'd prefer), it should go to the first (unnamed) section or to a separate `Notes` section (see https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections).

sfs/mono/source.py
```diff
@@ -79,6 +79,54 @@ def point_velocity(omega, x0, n0, grid, c=None):
...
+        plt.title("Dipole Point Source at {} m".format(x0))
+
+    """
+    k = util.wavenumber(omega, c)
+    x0 = util.asarray_1d(x0)
+    n0 = util.asarray_1d(n0)
+    grid = util.XyzComponents(grid)
+
+    r = np.linalg.norm(grid - x0)
+    scalar = np.inner((grid - x0), n0)
```
mgeier: `grid - x0` is repeated.
I called this `offset` in some places, but there are other places where it's called `dx`.
We should probably use only one of them.

mgeier: I think `scalar` is a strange variable name anyway, but especially if it's actually not a scalar, as in this case.

sfs/mono/source.py
hagenw: Should I move it behind or before the Examples section. If you say "first unnamed section" it sounds like it should go to the very end, which is fine as in my opnion the equation should be directly above the actual code.
I'm asking, because in the case of the point source it looks like this at the moment:
```python
def point(omega, x0, n0, grid, c=None):
    """Point source.

    ::

                      1  e^(-j w/c |x-x0|)
        G(x-x0, w) = --- -----------------
                     4pi      |x-x0|

    Examples
    --------
    .. plot::
        :context: close-figs

        p = sfs.mono.source.point(omega, x0, None, grid)
        sfs.plot.soundfield(p, grid)
        plt.title("Point Source at {} m".format(x0))
```

sfs/mono/source.py
hagenw: OK, I called it offset. I'm not totally convinced by the name, but have not better idea at the moment.
The line with scalar is now completely removed.

sfs/mono/source.py
hagenw: Done with f149ca7

sfs/mono/source.py
hagenw: Done with f149ca7

sfs/mono/source.py
mgeier: With "first (unnamed) section" I actually meant what is called "Extended Summary" in the [numpy docs](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections).

This is before the first *named* section, which is typically `Parameters`.

The docstring to `mono.source.point()` is (like many others) currently work-in-progress.

We should decide where to put the equations ("Extended Summary" or `Notes` section) and then change this in all docstrings.

sfs/mono/source.py
mgeier: I meant you should change `numpy.ndarray` to `array_like`!
I was only talking about the line before my comment (line 93).

You added `tripple` (which should be `triple` BTW) at many places where it doesn't belong!

hagenw: It was not so easy to see which line you meant, I thought all of the lines marked in green ;)

sfs/mono/source.py
hagenw: As the ASCII formula is only intended to help you within your editor, is it possible to remove it completely from the docstring and place it as a comment direct above the code? If it should stay in the docstring, then I would prefer to move it to the last entry their.

sfs/mono/source.py
mgeier: I didn't mark them, it Github's fault!

sfs/mono/source.py
mgeier: The docstring is for the description of the behavior of the function, that's exactly where the equation should be.
Comments are for implementation-specific remarks.
Since Python is such a readable language, comments can be quite rare ...

You cannot move the equation to the last section, because that's `Examples`, where it doesn't belong.

hagenw: I still find python extremely unreadable for equations, but ok ;)

Then I would propose the following solution until we will come back to the ASCII equations and remove them later:

We will put them into a Note section which comes before the Examples section.

sfs/mono/source.py
```diff
@@ -79,6 +81,57 @@ def point_velocity(omega, x0, n0, grid, c=None):
...
+
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        n0 = [0, 1, 0]
+        p = sfs.mono.source.point_dipole(omega, x0, n0, grid)
+        sfs.plot.soundfield(p, grid)
+        plt.title("Dipole Point Source at {} m".format(x0))
```
mgeier: The docstring should end with an empty line

sfs/mono/source.py
```diff
@@ -79,6 +81,57 @@ def point_velocity(omega, x0, n0, grid, c=None):
...
+
+    Notes
+    -----
+    ::
+
+         d                1   / iw       1    \   (x-x0) n0
+        ---- G(x-x0,w) = --- | ----- + ------- | ----------- e^(-i w/c |x-x0|)
+        d ns             4pi  \  c     |x-x0| /   |x-x0|^2
+
+    Examples
```
mgeier: I've used the plural "Examples" only if there are more than 2 plots, but I'm not sure if that was such a good idea.

Probably we should use "Examples" in all cases, even if there is only 1 plot?

hagenw: I would say so, it is also called "Examples" here:
https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#sections

sfs/mono/source.py
hagenw: OK, I have corrected it.

sfs/mono/source.py
mgeier: You're right. So please ignore my comment and I will change that at a later time.

sfs/mono/source.py
hagenw: I thought if pep8 does not complain everything must be fine ;)
It's now corrected.

mgeier: It's not in PEP8, I think it's not even in [PEP257](https://www.python.org/dev/peps/pep-0257/)!

I don't know if it's mentioned explicitly anywhere, but it is consistently used in the examples of the [NumPy docstring guidelines](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) and of course in the NumPy implementation itself.

I think it's worth mentioning that there should be an empty line at the end of the docstring, but no empty line following it.
To me this sounds counter-intuitive but now that I'm used to it it looks quite natural.

sfs/mono/source.py
```diff
@@ -91,7 +145,7 @@ def point_modal(omega, x0, n0, grid, L, N=None, deltan=0, c=None):
     n0 : (3,) array_like
         Normal vector (direction) of source (only required for
         compatibility).
-    grid : triple of numpy.ndarray
+    grid : triple of array_like
```
mgeier: While you're at it, you could change the same thing for `point_modal_velocity()`

PR #12, opened 2016-05-13
Update LICENSE

hagenw:
In the Matlab version and on the homepage I used the phrase `SFS Toolbox Team` for the copyright statements and would propose to use the same in the Python version.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2014-2015 SFS Toolbox Team
+Copyright (c) 2014-2015 SFS Toolbox contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2014-2015 SFS Toolbox Developers
+Copyright (c) 2014-2015 SFS Toolbox Team
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


mgeier:
Is there a precedent for this?

hagenw:
Developers suggest more that you have worked on the code, but there are some people that work only on the documentation or maybe the web page. That's the reason why I would prefer Team instead.

mgeier:
So there is no precedent?

Both NumPy and SciPy use "Developers".

"Team" sounds very exclusive, and a bit like a company is behind the whole thing, "Developers" puts the focus on the individuals.

Also, currently there is nobody that only works on the documentation. And I don't think there will be anytime soon.
The web page http://sfs.readthedocs.io/ is generated from content written by the SFS developers, the content of the web site http://sfstoolbox.org/ is not generated from content of this repository.

Therefore: I'm very much against "Team". :-1:

If you don't like "Developers", what about "Contributors"?
It would still be nice to have a precedent, but I think I've seen "XYZ Contributors" somewhere before ...

hagenw:
Here are two precedents:
https://raw.githubusercontent.com/nodejs/node/master/LICENSE
https://raw.githubusercontent.com/mockito/mockito/master/LICENSE
It seems to be that they use the lower case version "contributors".

I liked with "Team" that it is short, but I can understand your complains.
I changed the commit in this pull request to use "contributors" which I still find a little bit better than "Developers".

mgeier:
JavaScript and Java projects are *definitely not* valid sources for inspiration with regards to anything even remotely related to style or taste!

mgeier:
Those are some examples I found:

* NumPy Developers
* SciPy Developers
* IPython Development Team
* Jupyter Development Team
* Sphinx team
* authors of Vispy
* Kivy Team and other contributors

I still don't see any reason to change what we currently use.

hagenw:
One of the reasons is that we use something different on the web page and the Matlab version.
But as the easiest way seems to be to change it there, I will switch to `.SFS Toolbox Developers` there.

BTW, you should update the year in the LICENSE file of sfs-python.

mgeier:
An interesting discussion from the Jupyter project: https://groups.google.com/forum/#!msg/jupyter/vZpWgw8zKdc/2P27jQlaBAAJ

It looks like they'll change their copyright notice to "Project Jupyter Contributors".

PR #18, opened 2016-11-24
spherical Bessel functions

### LINKED ISSUE #7: Division by zero in NFC-HOA 2.5D plane driving function
Start with these settings

``` python
X = [-1.75, 1.75]
Y = [-1.75, 1.75]
Z = 0
nk = [0, -1, 0]
omega = 2 * np.pi * 1000
R0 = 1.5
```

First, try this one, this should still work:

``` python
>>> x0, n0, a0 = sfs.array.circular(100, R0)
>>> d = sfs.mono.drivingfunction.nfchoa_25d_plane(omega, x0, R0, nk)
```

Now increase the number of secondary sources and we will get a division by zero:

``` python
>>> x0, n0, a0 = sfs.array.circular(1000, R0)
>>> d = sfs.mono.drivingfunction.nfchoa_25d_plane(omega, x0, R0, nk)
/home/hagen/git/sfs-python/sfs/mono/drivingfunction.py:229: RuntimeWarning: divide by zero encountered in true_divide
  a = 1 / _sph_hn2(M, k * r0)
/home/hagen/git/sfs-python/sfs/mono/drivingfunction.py:229: RuntimeWarning: invalid value encountered in true_divide
  a = 1 / _sph_hn2(M, k * r0)
```

trettberg:
~~fix #7~~
* ~~HOA driving functions get an optional parameter 'Nmax' for maximum order~~
* (switched to different library functions for spherical Bessel functions)

`max_order` parameter is adressed in #21

### DIFF
--- a/sfs/mono/drivingfunction.py
+++ b/sfs/mono/drivingfunction.py
@@ -235,7 +235,7 @@ def source_selection_all(N):
     return np.ones(N, dtype=bool)
 
 
-def nfchoa_2d_plane(omega, x0, r0, n=[0, 1, 0], c=None):
+def nfchoa_2d_plane(omega, x0, r0, n=[0, 1, 0], Nmax=None, c=None):
     r"""Plane wave by two-dimensional NFC-HOA.
 
     .. math::
@@ -254,14 +254,14 @@ def nfchoa_2d_plane(omega, x0, r0, n=[0, 1, 0], c=None):
     n = util.asarray_1d(n)
     phi, _, r = util.cart2sph(*n)
     phi0 = util.cart2sph(*x0.T)[0]
-    M = _max_order_circular_harmonics(len(x0))
+    M = _max_order_circular_harmonics(len(x0),Nmax)
     d = 0
     for m in range(-M, M + 1):
         d += 1j**-m / hankel2(m, k * r0) * np.exp(1j * m * (phi0 - phi))
     return -2j / (np.pi*r0) * d
 
 
-def nfchoa_25d_point(omega, x0, r0, xs, c=None):
+def nfchoa_25d_point(omega, x0, r0, xs, Nmax=None, c=None):
     r"""Point source by 2.5-dimensional NFC-HOA.
 
     .. math::
@@ -280,14 +280,14 @@ def nfchoa_25d_point(omega, x0, r0, xs, c=None):
     xs = util.asarray_1d(xs)
     phi, _, r = util.cart2sph(*xs)
     phi0 = util.cart2sph(*x0.T)[0]
-    M = _max_order_circular_harmonics(len(x0))
+    M = _max_order_circular_harmonics(len(x0),Nmax)
     d = 0
     for m in range(-M, M + 1):
         d += _spherical_hn2(abs(m), k*r) / _spherical_hn2(abs(m), k*r0)  * np.exp(1j * m * (phi0 - phi))
     return d / (2 * np.pi * r0)
 
 
-def nfchoa_25d_plane(omega, x0, r0, n=[0, 1, 0], c=None):
+def nfchoa_25d_plane(omega, x0, r0, n=[0, 1, 0], Nmax=None, c=None):
     r"""Plane wave by 2.5-dimensional NFC-HOA.
 
     .. math::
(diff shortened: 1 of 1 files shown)


mgeier:
How should we handle that `special.spherical_jn()` and `special.spherical_yn()` are not available in SciPy < 0.18.0?
We should either try to keep the old implementation (which has the problems shown in #7) as a fallback, or we could drop support for SciPy < 0.18, but then I think we should create a more meaningful error message on older SciPy versions.

sfs/mono/drivingfunction.py
```diff
@@ -280,16 +280,14 @@ def nfchoa_25d_point(omega, x0, r0, xs, c=None):
...
     phi, _, r = util.cart2sph(*xs)
     phi0 = util.cart2sph(*x0.T)[0]
-    M = _max_order_circular_harmonics(len(x0))
-    hr = _sph_hn2(M, k * r)
-    hr0 = _sph_hn2(M, k * r0)
+    M = _max_order_circular_harmonics(len(x0),Nmax)
     d = 0
     for m in range(-M, M + 1):
-        d += hr[abs(m)] / hr0[abs(m)] * np.exp(1j * m * (phi0 - phi))
+        d += _spherical_hn2(abs(m), k*r) / _spherical_hn2(abs(m), k*r0)  * np.exp(1j * m * (phi0 - phi))
```
mgeier: Line too long, you should check for PEP 8.

sfs/mono/drivingfunction.py
```diff
@@ -308,11 +306,10 @@ def nfchoa_25d_plane(omega, x0, r0, n=[0, 1, 0], c=None):
     n = util.asarray_1d(n)
     phi, _, r = util.cart2sph(*n)
     phi0 = util.cart2sph(*x0.T)[0]
-    M = _max_order_circular_harmonics(len(x0))
-    hn2 = _sph_hn2(M, k * r0)
+    M = _max_order_circular_harmonics(len(x0),Nmax)
     d = 0
     for m in range(-M, M + 1):
-        d += 1j**-abs(m) / (k * hn2[abs(m)]) * np.exp(1j * m * (phi0 - phi))
+        d += 1j**-abs(m) / (k * _spherical_hn2(abs(m), k*r0)) * np.exp(1j * m * (phi0 - phi))
```
mgeier: Line too long.

sfs/mono/drivingfunction.py
```diff
@@ -254,14 +254,14 @@ def nfchoa_2d_plane(omega, x0, r0, n=[0, 1, 0], c=None):
     n = util.asarray_1d(n)
     phi, _, r = util.cart2sph(*n)
     phi0 = util.cart2sph(*x0.T)[0]
-    M = _max_order_circular_harmonics(len(x0))
+    M = _max_order_circular_harmonics(len(x0),Nmax)
```
mgeier: Missing space.

sfs/mono/drivingfunction.py
```diff
@@ -280,16 +280,14 @@ def nfchoa_25d_point(omega, x0, r0, xs, c=None):
     xs = util.asarray_1d(xs)
     phi, _, r = util.cart2sph(*xs)
     phi0 = util.cart2sph(*x0.T)[0]
-    M = _max_order_circular_harmonics(len(x0))
-    hr = _sph_hn2(M, k * r)
-    hr0 = _sph_hn2(M, k * r0)
+    M = _max_order_circular_harmonics(len(x0),Nmax)
```
mgeier: Missing space.

mgeier:
I don't like the name `Nmax`, especially since it's right next to `n`, which is something completely different.
What about `maxorder` or `max_order` or something like that?

sfs/mono/drivingfunction.py
```diff
@@ -667,12 +664,11 @@ def esa_edge_dipole_2d_line(omega, x0, xs, alpha=3/2*np.pi, Nc=None, c=None):
...
-    jn, jnp, yn, ynp = sph_jnyn(n, z)
-    return jn - 1j * yn
+    return spherical_jn(n,z) - 1j * spherical_yn(n,z)
 
 
-def _max_order_circular_harmonics(N):
+def _max_order_circular_harmonics(N,Nmax):
     """Compute order of 2D HOA."""
-    return N // 2 if N % 2 == 0 else (N - 1) // 2
+    return Nmax if Nmax else N // 2 if N % 2 == 0 else (N - 1) // 2
```
mgeier: Zeroth order is probably not something many people would want to use, but your logic should still account for it, right?
I think you should check `if Nmax is not None`.

trettberg:
> How should we handle that `special.spherical_jn()` and `special.spherical_yn()` are not available in SciPy < 0.18.0?

The old `sph_jnyn` has a different signature. IMHO supporting both versions is not worth the trouble.

I'd suggest to stay with old version for now.
At some point in time, new features (3D-NFCHOA, directional sources) might justify to drop pre-0.18.0 support.


> I don't like the name `Nmax`

Maybe `max_order` could be used consistently for all functions that deal with truncated modal expansions.

mgeier:
What about splitting this into two PRs? The first about `max_order` can be merged right now, and for the other one we can wait some time.

trettberg:
Of course you're right. This should have been two separate PRs to begin with.
#21 deals with HOA order

PR #21, opened 2016-11-30
Add parameter max_order to NFCHOA driving functions

### LINKED ISSUE #7: Division by zero in NFC-HOA 2.5D plane driving function

trettberg:
was previously #18, fixes #7

### DIFF
--- a/sfs/mono/drivingfunction.py
+++ b/sfs/mono/drivingfunction.py
@@ -235,7 +235,7 @@ def source_selection_all(N):
     return np.ones(N, dtype=bool)
 
 
-def nfchoa_2d_plane(omega, x0, r0, n=[0, 1, 0], c=None):
+def nfchoa_2d_plane(omega, x0, r0, n=[0, 1, 0], max_order=None, c=None):
     r"""Plane wave by two-dimensional NFC-HOA.
 
     .. math::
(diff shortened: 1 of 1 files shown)


sfs/mono/drivingfunction.py
```diff
@@ -673,6 +673,7 @@ def _sph_hn2(n, z):
     return jn - 1j * yn
 
 
-def _max_order_circular_harmonics(N):
+def _max_order_circular_harmonics(N, max_order):
     """Compute order of 2D HOA."""
-    return N // 2 if N % 2 == 0 else (N - 1) // 2
+    return (max_order if max_order is not None
+            else N // 2 if N % 2 == 0 else (N - 1) // 2)
```
mgeier: I really like if/else expressions, but as soon as they need more than one lines, they lose a bit of their appeal ...

How about this:

```python
    if max_order is not None:
        return max_order
    return ...
```

And am I crazy or is the last line exactly equivalent to simply `N // 2`?

If yes, we can go back to the nice one-liner:

```python
return max_order if max_order is not None else N // 2
```

Or even better:

```python
return N // 2 if max_order is None else max_order
```

I overlooked that when I changed it in 39787214e6808ecaa2910ec2159925ef98fbe0de.

trettberg: Nice one! Behaviour is equivalent for integer `N` but differs for fractional. That was probably not intended so I changed it to suggestion.

PR #23, opened 2016-12-01
Add time based 2.5D WFS plane wave driving function.

chris-hld:
Major improvements to time domain implementation.

### DIFF
--- a/sfs/time/drivingfunction.py
+++ b/sfs/time/drivingfunction.py
@@ -126,7 +126,7 @@ def wfs_25d_point(x0, n0, xs, xref=[0, 0, 0], c=None):
     ds = x0 - xs
     r = np.linalg.norm(ds, axis=1)
     delays = r/c
-    weights = g0/(2 * np.pi) * inner1d(ds, n0) / r**(3/2)
+    weights = g0 * inner1d(ds, n0) / (2 * np.pi * r**(3/2))
     return delays, weights
 
 
@@ -150,7 +150,7 @@ def driving_signals(delays, weights, signal, fs=None):
     Returns
     -------
     driving_signals : (N, C) numpy.ndarray
-        Driving signals (with N samples) per channel C.
+        Driving signal per channel (column represents channel).
     t_offset : float
         Simulation point in time offset (seconds).
 
@@ -162,10 +162,11 @@ def driving_signals(delays, weights, signal, fs=None):
 
 
 def apply_delays(signal, delays, fs=None):
-    """Apply Delays for every channel.
+    """Apply delays for every channel.
 
-    Delay in seconds, no fractional delay. An absolute time offset is applied
-    and returned.
+    A mono input signal gets delayed for each channel individually.
+    The simultation point in time is shifted by the smallest delay provided,
+    which allows negative delays as well.
 
     Parameters
     ----------
@@ -179,7 +180,7 @@ def apply_delays(signal, delays, fs=None):
     Returns
     -------
     out : (N, C) numpy.ndarray
-        Output signals (with N samples) per channel C.
+        Output signals (column represents channel).
     t_offset : float
         Simulation point in time offset (seconds).
 
--- a/sfs/time/drivingfunction.py
+++ b/sfs/time/drivingfunction.py
@@ -63,7 +63,7 @@ def wfs_25d_plane(x0, n0, n=[0, 1, 0], xref=[0, 0, 0], c=None):
     n = util.asarray_1d(n)
     xref = util.asarray_1d(xref)
     g0 = np.sqrt(2 * np.pi * np.linalg.norm(xref - x0, axis=1))
-    delays = inner1d(n, x0)/c
+    delays = inner1d(n, x0) / c
     weights = 2 * g0 * inner1d(n, n0)
     return delays, weights
 
@@ -124,7 +124,7 @@ def wfs_25d_point(x0, n0, xs, xref=[0, 0, 0], c=None):
     xref = util.asarray_1d(xref)
     g0 = np.sqrt(2 * np.pi * np.linalg.norm(xref - x0, axis=1))
     ds = x0 - xs
-    r = np.linalg.norm(x0 - xs, axis=1)
+    r = np.linalg.norm(ds, axis=1)
     delays = r/c
     weights = g0/(2 * np.pi) * inner1d(ds, n0) / r**(3/2)
     return delays, weights
@@ -133,7 +133,8 @@ def wfs_25d_point(x0, n0, xs, xref=[0, 0, 0], c=None):
 def driving_signals(delays, weights, signal, fs=None):
     """Get driving signals per secondary source.
 
-    Returned signal is the delayed and weighted input signal per channel.
+    Returned signals are the delayed and weighted mono input signal (with N
+    samples) per channel (C).
 
     Parameters
     ----------
@@ -144,12 +145,12 @@ def driving_signals(delays, weights, signal, fs=None):
     signal : (N,) array_like
         Excitation signal (mono) which gets weighted and delayed.
     fs: int, optional
-        Sampling frequency in hertz.
+        Sampling frequency in Hertz.
 
     Returns
     -------
-    (N, C) numpy.ndarray
-        N is signal samples(t) per channel C, delayed and weighted.
+    driving_signals : (N, C) numpy.ndarray
+        Driving signals (with N samples) per channel C.
     t_offset : float
         Simulation point in time offset (seconds).
 
@@ -163,21 +164,22 @@ def driving_signals(delays, weights, signal, fs=None):
 def apply_delays(signal, delays, fs=None):
     """Apply Delays for every channel.
 
-    Delay in seconds, no fractional delay. An absolute time offset is applied.
+    Delay in seconds, no fractional delay. An absolute time offset is applied
+    and returned.
 
     Parameters
     ----------
     signal : (N,) array_like
-        Excitation signal (mono) which gets weighted and delayed.
-    delay : (C,) array_like, in seconds
-        delay in seconds for each channel, negative values allowed.
+        Mono excitation signal (with N samples) which gets delayed.
+    delays : (C,) array_like
+        Delay in seconds for each channel (C), negative values allowed.
     fs: int, optional
-        Sampling frequency in hertz.
+        Sampling frequency in Hertz.
 
     Returns
     -------
     out : (N, C) numpy.ndarray
-        N is signal samples(t) per channel C.
+        Output signals (with N samples) per channel C.
     t_offset : float
         Simulation point in time offset (seconds).
 
@@ -187,11 +189,10 @@ def apply_delays(signal, delays, fs=None):
     signal = util.asarray_1d(signal)
     delays = util.asarray_1d(delays)
 
-    # convert to samples
     delays_samples = np.rint(fs * delays).astype(int)
     offset_samples = delays_samples.min()
     delays_samples -= offset_samples
     out = np.zeros((delays_samples.max() + len(signal), len(delays_samples)))
     for channel, cdelay in enumerate(delays_samples):
-        out[cdelay: cdelay + len(signal), channel] = signal
+        out[cdelay:cdelay + len(signal), channel] = signal
     return out, offset_samples / fs
--- a/sfs/time/drivingfunction.py
+++ b/sfs/time/drivingfunction.py
@@ -133,8 +133,8 @@ def wfs_25d_point(x0, n0, xs, xref=[0, 0, 0], c=None):
 def driving_signals(delays, weights, signal, fs=None):
     """Get driving signals per secondary source.
 
-    Returned signals are the delayed and weighted mono input signal (with N
-    samples) per channel (C).
+    Returned signals are the delayed and weighted mono input signal
+    (with N samples) per channel (C).
 
     Parameters
     ----------
@@ -164,8 +164,8 @@ def driving_signals(delays, weights, signal, fs=None):
 def apply_delays(signal, delays, fs=None):
     """Apply delays for every channel.
 
-    A mono input signal gets delayed for each channel individually.
-    The simultation point in time is shifted by the smallest delay provided,
+    A mono input signal gets delayed for each channel individually. The
+    simultation point in time is shifted by the smallest delay provided,
     which allows negative delays as well.
 
     Parameters
--- a/sfs/time/drivingfunction.py
+++ b/sfs/time/drivingfunction.py
@@ -1,64 +1,197 @@
-"""Compute time based driving functions for various systems."""
+"""Compute time based driving functions for various systems.
 
+.. include:: math-definitions.rst
+
+"""
+from __future__ import division
 import numpy as np
 from numpy.core.umath_tests import inner1d  # element-wise inner product
 from .. import defs
 from .. import util
 
 
-def wfs_25d_ps(xs, x0, nx0, xref=[0, 0, 0], fs=None, c=None):
-    """Point source by 2.5-dimensional WFS.
+def wfs_25d_plane(x0, n0, n=[0, 1, 0], xref=[0, 0, 0], c=None):
+    r"""Plane wave model by 2.5-dimensional WFS.
 
-    ::
+    Parameters
+    ----------
+    x0 : (N, 3) array_like
+        Sequence of secondary source positions.
+    n0 : (N, 3) array_like
+        Sequence of secondary source orientations.
+    n : (3,) array_like, optional
+        Normal vector (propagation direction) of synthesized plane wave.
+    xref : (3,) array_like, optional
+        Reference position
+    c : float, optional
+        Speed of sound
 
-         2.5D correction factor
-                ______________
-         g0 = \| 2pi |xref-x0|
+    Returns
+    -------
+    delays : (N,) numpy.ndarray
+        Delays of secondary sources in seconds.
+    weights: (N,) numpy.ndarray
+        Weights of secondary sources.
 
+    Notes
+    -----
+    2.5D correction factor
 
-         d_2.5D using a point source as source model
+    .. math::
 
-                               g0  (x0-xs) nx0
-         d_2.5D(x0,t) = h(t) * --- ------------- delta(t-|x0-xs|/c)
-                               2pi |x0-xs|^(3/2)
+        g_0 = \sqrt{2 \pi |x_\mathrm{ref} - x_0|}
 
-         see Wierstorf et al. (2015), eq.(#d:wfs:ps:2.5D)
+    d using a plane wave as source model
+
+    .. math::
+
+        d_{2.5D}(x_0,t) = h(t)
+        2 g_0 \scalarprod{n}{n_0}
+        \dirac{t - \frac{1}{c} \scalarprod{n}{x_0}}
+
+    with wfs(2.5D) prefilter h(t), which is not implemented yet.
+
+    References
+    ----------
+    See http://sfstoolbox.org/en/latest/#equation-d.wfs.pw.2.5D
 
     """
     if c is None:
         c = defs.c
-    if fs is None:
-        fs = defs.fs
     x0 = util.asarray_of_rows(x0)
-    nx0 = util.asarray_of_rows(nx0)
-    xs = util.asarray_1d(xs)
-    g0 = np.sqrt(2*np.pi*np.linalg.norm(xref-x0, axis=1))
-    ds = x0 - xs
-    r = np.linalg.norm(x0-xs, axis=1)
-    delay = 1/c * r
-    weight = g0/(2*np.pi) * inner1d(ds, nx0) / r**(3/2)
-    sig = np.ones(len(x0))  # dirac
-    line = delayline(sig, delay*fs, weight)
-    return delay, weight, line
+    n0 = util.asarray_of_rows(n0)
+    n = util.asarray_1d(n)
+    xref = util.asarray_1d(xref)
+    g0 = np.sqrt(2 * np.pi * np.linalg.norm(xref - x0, axis=1))
+    delays = inner1d(n, x0)/c
+    weights = 2 * g0 * inner1d(n, n0)
+    return delays, weights
 
 
-def delayline(sig, delay, weight):
-    """Delayline with weights, no fractional delay yet.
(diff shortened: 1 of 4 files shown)


sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,197 @@
...
+        Amplitude weighting factor for each channel.
+    signal : (N,) array_like
+        Excitation signal (mono) which gets weighted and delayed.
+    fs: int, optional
+        Sampling frequency in hertz.
+
+    Returns
+    -------
+    (N, C) numpy.ndarray
+        N is signal samples(t) per channel C, delayed and weighted.
```
mgeier: I don't understand this sentence. What does "t" mean?

I think you should move the explanation of C and N up, because those are used before.

sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,197 @@
...
-def delayline(sig, delay, weight):
-    """Delayline with weights, no fractional delay yet.
+def apply_delays(signal, delays, fs=None):
+    """Apply Delays for every channel.
 
-    Delay in samples.
-    sig (m x n) where columns n are channels and rows m the corresponding
-    signals.
-    Weight is array with weighting for each channel.
+    Delay in seconds, no fractional delay. An absolute time offset is applied.
```
mgeier: You mean "... is returned."?

sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,197 @@
...
-    delay = np.rint(delay).astype(int)  # no fractional delay, samples
-    extention = np.max(delay)  # avoid cutting signal
-    channels = len(weight)  # extract no. of channels
-    sig_length = sig.ndim
-    out = np.zeros([sig_length+extention, channels])  # create shape
+    Parameters
+    ----------
+    signal : (N,) array_like
+        Excitation signal (mono) which gets weighted and delayed.
+    delay : (C,) array_like, in seconds
```
mgeier: `delays`

sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,197 @@
...
-    """
-    delay = np.rint(delay).astype(int)  # no fractional delay, samples
-    extention = np.max(delay)  # avoid cutting signal
-    channels = len(weight)  # extract no. of channels
-    sig_length = sig.ndim
-    out = np.zeros([sig_length+extention, channels])  # create shape
+    Parameters
+    ----------
+    signal : (N,) array_like
+        Excitation signal (mono) which gets weighted and delayed.
```
mgeier: There doesn't seem to be any weighting involved, right?

sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,197 @@
...
+    Parameters
+    ----------
+    delays : (C,) array_like
+        Delay in seconds for each channel, negative values allowed.
+    weights : (C,) array_like
+        Amplitude weighting factor for each channel.
+    signal : (N,) array_like
+        Excitation signal (mono) which gets weighted and delayed.
+    fs: int, optional
+        Sampling frequency in hertz.
```
mgeier: Hertz

sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,197 @@
...
-    sig_length = sig.ndim
-    out = np.zeros([sig_length+extention, channels])  # create shape
+    Parameters
+    ----------
+    signal : (N,) array_like
+        Excitation signal (mono) which gets weighted and delayed.
+    delay : (C,) array_like, in seconds
+        delay in seconds for each channel, negative values allowed.
+    fs: int, optional
+        Sampling frequency in hertz.
```
mgeier: Hertz

sfs/time/soundfield.py
```diff
@@ -1,48 +1,57 @@
...
+    t : float
+        Simulation point in time (seconds).
+    grid : triple of array_like
+        The grid that is used for the sound field calculations.
+        See `sfs.util.xyz_grid()`.
+    source: function, optional
+        Source type is a function, returning scalar field.
+        For default, see `sfs.time.source.point()`
+    fs: int, optional
+        Sampling frequency in hertz.
```
mgeier: Hertz

sfs/time/drivingfunction.py
mgeier: superfluous "in seconds"

sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,197 @@
...
         c = defs.c
-    if fs is None:
-        fs = defs.fs
     x0 = util.asarray_of_rows(x0)
-    nx0 = util.asarray_of_rows(nx0)
+    n0 = util.asarray_of_rows(n0)
+    n = util.asarray_1d(n)
+    xref = util.asarray_1d(xref)
+    g0 = np.sqrt(2 * np.pi * np.linalg.norm(xref - x0, axis=1))
+    delays = inner1d(n, x0)/c
```
mgeier: This would be nicer with spaces around the `/`

sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,197 @@
...
     ds = x0 - xs
-    r = np.linalg.norm(x0-xs, axis=1)
-    delay = 1/c * r
-    weight = g0/(2*np.pi) * inner1d(ds, nx0) / r**(3/2)
-    sig = np.ones(len(x0))  # dirac
-    line = delayline(sig, delay*fs, weight)
-    return delay, weight, line
+    r = np.linalg.norm(x0 - xs, axis=1)
+    delays = r/c
+    weights = g0/(2 * np.pi) * inner1d(ds, n0) / r**(3/2)
```
mgeier: You could use the same numerator/denominator as in the docstring, the two fractions here look a bit confusing, don't they?

Or is there a reason for writing it like that?

sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,197 @@
...
+    signal = util.asarray_1d(signal)
+    delays = util.asarray_1d(delays)
+
+    # convert to samples
+    delays_samples = np.rint(fs * delays).astype(int)
+    offset_samples = delays_samples.min()
+    delays_samples -= offset_samples
+    out = np.zeros((delays_samples.max() + len(signal), len(delays_samples)))
+    for channel, cdelay in enumerate(delays_samples):
+        out[cdelay: cdelay + len(signal), channel] = signal
```
mgeier: superfluous space after colon

sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,197 @@
...
+        Simulation point in time offset (seconds).
 
-    return out
+    """
+    if fs is None:
+        fs = defs.fs
+    signal = util.asarray_1d(signal)
+    delays = util.asarray_1d(delays)
+
+    # convert to samples
```
mgeier: Since the word `samples` is very frequently used in the following few lines, is this comment really necessary?

sfs/time/soundfield.py
```diff
@@ -1,48 +1,57 @@
...
+
+
+def p_array(x0, d, channel_weights, t, grid, source=point, fs=None, c=None):
+    """Compute sound field for an array of secondary sources for a given time.
+
+    Parameters
+    ----------
+    x0 : (N, 3) array_like
+        Sequence of secondary source positions.
+    d : (N, C) numpy.ndarray
```
mgeier: `array_like`

sfs/time/soundfield.py
```diff
@@ -1,48 +1,57 @@
...
+        For default, see `sfs.time.source.point()`
+    fs: int, optional
+        Sampling frequency in hertz.
+    c : float, optional
+        Speed of sound.
+
+    Returns
+    -------
+    numpy.ndarray
+        Pressure at grid positions.
```
mgeier: Sound pressure

sfs/time/soundfield.py
```diff
@@ -1,48 +1,57 @@
...
+    channel_weights : (C,) array_like
+        Additional weights applied during integration, e.g. source tapering.
+    t : float
+        Simulation point in time (seconds).
+    grid : triple of array_like
+        The grid that is used for the sound field calculations.
+        See `sfs.util.xyz_grid()`.
+    source: function, optional
+        Source type is a function, returning scalar field.
+        For default, see `sfs.time.source.point()`
```
mgeier: .

sfs/time/soundfield.py
```diff
@@ -1,48 +1,57 @@
...
     """
     if fs is None:
         fs = defs.fs
-    channel_weight = util.asarray_1d(channel_weight)
-    if len(channel_weight) != len(x0):
+    if c is None:
+        c = defs.c
+    x0 = util.asarray_of_rows(x0)
+    channel_weights = util.asarray_1d(channel_weights)
+    if np.size(channel_weights, 0) != np.size(x0, 0):
```
mgeier: Every time someone uses `np.size()`, a fairy dies. To avoid this, you should use `len(channel_weights)` and `len(x0)`.

mgeier: The number of columns of `d` must also be the same.

And before killing another fairy, please use `d.shape[1]` or `len(d.T)`, I'm not sure which one is better.

sfs/time/soundfield.py
```diff
@@ -1,48 +1,57 @@
...
+        Pressure at grid positions.
 
     """
     if fs is None:
         fs = defs.fs
-    channel_weight = util.asarray_1d(channel_weight)
-    if len(channel_weight) != len(x0):
+    if c is None:
+        c = defs.c
+    x0 = util.asarray_of_rows(x0)
```
mgeier: You should make sure that `d` is an array:

```python
d = np.asarray(d)
```

Probably now is the time to create a function `asarray_of_columns()`?

sfs/time/source.py
```diff
@@ -2,30 +2,58 @@
...
 from .. import util
 from .. import defs
 
 
-def greens_function_ps(xs, grid, t, fs=None, c=None):
-    """Source model for a point source: 3D Green's function.
+def point(xs, signal, t, grid, fs=None, c=None):
+    r"""Source model for a point source: 3D Green's function.
+
+    Calculates the scalar pressure field for a given point in time, evoked by
```
mgeier: sound pressure

sfs/time/source.py
```diff
@@ -2,30 +2,58 @@
...
+        See `sfs.util.xyz_grid()`.
+    fs: int, optional
+        Sampling frequency in hertz.
+    c : float, optional
+        Speed of sound.
+
+    Returns
+    -------
+    numpy.ndarray
+        Scalar pressure field, evaluated at positions given by *grid*.
```
mgeier: sound pressure

sfs/time/source.py
```diff
@@ -2,30 +2,58 @@
...
-                     4pi |x-xs|
+        g(x-x_s,t) = \frac{1}{4 \pi |x - x_s|} \dirac{t - \frac{|x - x_s|}{c}}
 
     """
     xs = util.asarray_1d(xs)
     grid = util.as_xyz_components(grid)
     if c is None:
         c = defs.c
-    if fs is None:
-        fs = defs.fs
```
mgeier: Why did you remove that?

sfs/time/source.py
```diff
@@ -2,30 +2,58 @@
...
+    -----
+    .. math::
 
-                         1
-         g(x-xs,t) = ---------- delta(t - |x-xs|/c)
-                     4pi |x-xs|
+        g(x-x_s,t) = \frac{1}{4 \pi |x - x_s|} \dirac{t - \frac{|x - x_s|}{c}}
 
     """
     xs = util.asarray_1d(xs)
```
mgeier: You should ensure that `signal` is mono.

sfs/time/source.py
mgeier: You can of course remove it, but then you should also remove the default value `fs=None` above.

sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,197 @@
...
+    weights : (C,) array_like
+        Amplitude weighting factor for each channel.
+    signal : (N,) array_like
+        Excitation signal (mono) which gets weighted and delayed.
+    fs: int, optional
+        Sampling frequency in hertz.
+
+    Returns
+    -------
+    (N, C) numpy.ndarray
```
chris-hld: In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/drivingfunction.py
chris-hld: Also applied, isn't it?

sfs/time/drivingfunction.py
chris-hld: In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/drivingfunction.py
chris-hld: True. In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/drivingfunction.py
chris-hld: In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/drivingfunction.py
chris-hld: In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/soundfield.py
chris-hld: In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/source.py
```diff
@@ -2,30 +2,58 @@
...
+        Position of source in cartesian coordinates.
+    signal : (N,) array_like
+        Excitation signal.
+    t : float
+        Observed point in time.
+    grid : triple of array_like
+        The grid that is used for the sound field calculations.
+        See `sfs.util.xyz_grid()`.
+    fs: int, optional
+        Sampling frequency in hertz.
```
chris-hld: In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/drivingfunction.py
chris-hld: In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,197 @@
...
+    xref = util.asarray_1d(xref)
+    g0 = np.sqrt(2 * np.pi * np.linalg.norm(xref - x0, axis=1))
     ds = x0 - xs
-    r = np.linalg.norm(x0-xs, axis=1)
-    delay = 1/c * r
-    weight = g0/(2*np.pi) * inner1d(ds, nx0) / r**(3/2)
-    sig = np.ones(len(x0))  # dirac
-    line = delayline(sig, delay*fs, weight)
-    return delay, weight, line
+    r = np.linalg.norm(x0 - xs, axis=1)
```
chris-hld: Absolutely. In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/drivingfunction.py
chris-hld: I thought it's way nicer to read it like that, because of the bit bulky linalg.norm() etc..

sfs/time/drivingfunction.py
chris-hld: In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/drivingfunction.py
chris-hld: No. commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/soundfield.py
chris-hld: In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/soundfield.py
chris-hld: In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/source.py
chris-hld: In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

sfs/time/source.py
chris-hld: In commit dc23fc443a07d4103b36d84525725a9a9d0be6e6

chris-hld:
Thank you for that very detailed review!! I'm about to finish the remaining updates.

sfs/time/soundfield.py
chris-hld: This causes problems for only one channel, doesn't it?

sfs/time/source.py
chris-hld: In commit 8a77179efe033fc4a25c171dd675fc2be2a7b334

sfs/time/source.py
chris-hld: in commit 8a77179efe033fc4a25c171dd675fc2be2a7b334

sfs/time/soundfield.py
chris-hld: Yes, we could add that!

sfs/time/soundfield.py
chris-hld: in commit 8a77179efe033fc4a25c171dd675fc2be2a7b334

sfs/time/soundfield.py
chris-hld: in commit 8a77179efe033fc4a25c171dd675fc2be2a7b334

sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,198 @@
...
+        Amplitude weighting factor for each channel.
+    signal : (N,) array_like
+        Excitation signal (mono) which gets weighted and delayed.
+    fs: int, optional
+        Sampling frequency in Hertz.
+
+    Returns
+    -------
+    driving_signals : (N, C) numpy.ndarray
+        Driving signals (with N samples) per channel C.
```
mgeier: I think you should extend the parentheses until the end of the line, or remove them completely (which might be even better).

sfs/time/drivingfunction.py
mgeier: I'm not sure. Aren't you talking about the return parameter `delay`?
As far as I understand it, this is not *applied*. If anything, it is *removed* (or *subtracted*) from all the delays in `delays`.
But I'm not a native speaker, so I might be completely wrong.

sfs/time/drivingfunction.py
mgeier: Looking at it again, I'm not so sure anymore ... but something seems strange here.

Probably it's not necessary to mention `N`, it isn't mentioned above, either.

Also, the plural seems to be inappropriate. I think it's either one signal per channel or one multichannel signal with `C` channels or `C` signals, but not multiple signals per channel.

Same for `apply_delays()` below.

sfs/time/drivingfunction.py
```diff
@@ -1,64 +1,198 @@
...
+        Mono excitation signal (with N samples) which gets delayed.
+    delays : (C,) array_like
+        Delay in seconds for each channel (C), negative values allowed.
+    fs: int, optional
+        Sampling frequency in Hertz.
+
+    Returns
+    -------
+    out : (N, C) numpy.ndarray
+        Output signals (with N samples) per channel C.
```
mgeier: See above.

sfs/time/soundfield.py
mgeier: Yes, it should indeed be two-dimensional.

sfs/time/soundfield.py
mgeier: Do you want to add this to this PR our shall we do it afterwards?

sfs/time/soundfield.py
mgeier: No need to use indexing here. `len(x0)` is the length of x0, regardless if its one- or two-dimensional.

But if we had a `asarray_of_columns()` function, this would guarantee that the array is always two-dimensional, even for mono input.

sfs/time/soundfield.py
```diff
@@ -1,48 +1,57 @@
...
-    # zero padding and reverse time signals
-    d_line_inverted = np.flipud(np.row_stack((zero_fill, d_line, zero_fill)))
-
+    if c is None:
+        c = defs.c
+    x0 = util.asarray_of_rows(x0)
+    channel_weights = util.asarray_1d(channel_weights)
+    d = np.asarray(d)
+    if (len(channel_weights) != x0.shape[0]) or (
+            len(channel_weights) != d.shape[1]):
```
mgeier: No need for `or` here, you can just chain the comparisons. Yay Python!

And `x0.shape[0]` is equivalent to `len(x0)`. I would prefer latter.

mgeier:
Thanks for the changes! I hope I could clear up the few remaining points.

I have one more thing: PEP 8 tells us to limit docstring lines to 72 characters:
https://www.python.org/dev/peps/pep-0008/#maximum-line-length

Would be nice if you could also do this, right now some lines are longer.
Of course there may be some exceptions, but in most cases I think it would be good to limit the docstring line length to 72.

sfs/time/soundfield.py
chris-hld: I'd like to do this afterwards, because I think I'll need help with that.

mgeier: Sure, no problem.

Just for the record, it could be very similar to `asarray_of_rows()`, there's really only one character moved to a different place:

```python
def asarray_of_columns(a, **kwargs):
    result = np.squeeze(np.asarray(a, **kwargs))
    if result.ndim == 1:
        result = result.reshape(-1, 1)
    return result
```

sfs/time/soundfield.py
chris-hld: Yay Python! In commit 6765559c5cc2065e08a2254e0e3d5413fb3edde7

sfs/time/drivingfunction.py
chris-hld: How do you like this: In 6765559c5cc2065e08a2254e0e3d5413fb3edde7

chris-hld:
I updated the remaining points, however, simple and short docstring explanations become more difficult to me... Feel free to comment further!

mgeier:
Writing good docstrings is indeed hard. But I think yours are already better than those in other submodules. We should continually try to improve all docstrings in the future.

But first, LGTM!

Thanks for the great work!

PR #24, opened 2016-12-11
Add signal class and helper function

mgeier:
For now, this is just a proof of concept.

Alternatives are very welcome.
And suggestions for improvements, too.

For now, there are no doc changes yet. I'll do that if we decide to go that way ...

### DIFF
--- a/examples/time_domain.py
+++ b/examples/time_domain.py
@@ -15,18 +15,18 @@ my_cmap = 'YlOrRd'
 N = 56  # number of secondary sources
 R = 1.5  # radius of spherical/circular array
 x0, n0, a0 = sfs.array.circular(N, R)  # get secondary source positions
+fs = 44100  # sampling rate
 
-# create dirac
-signal = [1]
+# unit impulse
+signal = [1], fs
 
 # POINT SOURCE
 xs = [2, 2, 0]  # position of virtual source
 t = 0.008
 # compute driving signals
 d_delay, d_weight = sfs.time.drivingfunction.wfs_25d_point(x0, n0, xs)
-d, t_offset = sfs.time.drivingfunction.driving_signals(d_delay, d_weight,
-                                                       signal)
-t -= t_offset
+d = sfs.time.drivingfunction.driving_signals(d_delay, d_weight, signal)
+
 # test soundfield
 a = sfs.mono.drivingfunction.source_selection_point(n0, x0, xs)
 twin = sfs.tapering.tukey(a, .3)
@@ -48,9 +48,8 @@ t = -0.001
 
 # compute driving signals
 d_delay, d_weight = sfs.time.drivingfunction.wfs_25d_plane(x0, n0, npw)
-d, t_offset = sfs.time.drivingfunction.driving_signals(d_delay,
-                                                       d_weight, signal)
-t -= t_offset
+d = sfs.time.drivingfunction.driving_signals(d_delay, d_weight, signal)
+
 # test soundfield
 a = sfs.mono.drivingfunction.source_selection_plane(n0, npw)
 twin = sfs.tapering.tukey(a, .3)
--- a/sfs/defs.py
+++ b/sfs/defs.py
@@ -8,6 +8,3 @@ rho0 = 1.2250
 
 # tolerance used for secondary source selection
 selection_tolerance = 1e-6
-
-# sampling frequency
-fs = 44100
--- a/sfs/time/drivingfunction.py
+++ b/sfs/time/drivingfunction.py
@@ -130,7 +130,7 @@ def wfs_25d_point(x0, n0, xs, xref=[0, 0, 0], c=None):
     return delays, weights
 
 
-def driving_signals(delays, weights, signal, fs=None):
+def driving_signals(delays, weights, signal):
     """Get driving signals per secondary source.
 
     Returned signals are the delayed and weighted mono input signal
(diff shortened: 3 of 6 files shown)


trettberg:
What is supposed to happen when a function receives multiple signals with different sampling rates?

* raise an error?
* perform automatic resampling?

mgeier:
@trettberg Both behaviors are possible choices, we can decide once we actually have a function that takes more than one signal. We could even add an argument that specifies if resampling should be used or not.

trettberg:
My line of thought was something like:
If different sampling rates are not handled automatically, we don't have to store them with the signals.
The constant in `defs` might be enough?

mgeier:
@trettberg I agree with your edited comment. The sampling rate could be implicit (only handled by `defs`), but I think this has a high potential for errors.
I think it's best to make it explicit and obligatory. This is the case in this very PR.

However, a signal class could be created without explicit sampling rate and an explicit sampling rate could be used without a signal class.

As I said, IMHO the sampling rate should always be explicit, but there could still be a better alternative for the signal class proposed here.

Are there any counter-proposals?

trettberg:
From what i've tried so far, this works nicely for keeping track of processing delays.
I'm with it!

A suggestion for later:
IMHO this begs for overloading the class's arithmetic operators.
For example, adding driving signals for two virtual sources with `d1 + d2` should take care of time offsets (and sample rates).

mgeier:
I'm not sure if overloading the operators is the right way to go ... but I also don't know if it will be even needed, we'll see about that in the future.

I don't think it's a good idea to build too many features into the `Signal` class which would in the end lead to something like http://www.briansimulator.org/docs/reference-hears.html#sounds.

We should definitely consider using free functions instead of methods and overloaded operators.
If we use methods, we create a lock-in for our users, since they are forced to convert to a `Signal` object before they can do anything useful with it.
Overloaded operators might have to deal with some strange edge cases like:

```python
somedata = ...
sig1 = somedata, 44100
sig1 = sfs.util.as_signal(sig1)
sig2 = sig1 + ([1], 44100, 0.007)
```

I don't think this is the way we want to go, but it is definitely a possibility ....
Consider, OTOH, using a (hypothetical) free function:

```python
somedata = ...
sig1 = somedata, 44100
sig2 = sfs.util.add_signals(sig1, ([1], 44100, 0.007))
```

I think this is better, but it would probably be even better not having to add signals at all.

Since the main feature of the proposed signal class is the delay, we should probably rename it to something more specific?
What about `DelayedSignal` and `as_delayed_signal()`?
This should show that it is *not* a generic signal class (with many features) but rather a very specialized thing.

mgeier:
I switched to `DelayedSignal` and updated the documentation.

Any comments? Suggestions for improvements?

trettberg:
Superposition of signals is necessary e.g. to generate driving functions for data-based SFS. 
Acoustic room modeling (think image sources) is another application.

I do not suggest to stuff arbitrary functionality in the class but to get the math right:
For ease of illustration, assume all signals have 
* the same number of channels *C*, 
* a sampling rate of 1, 
* integer delays.

These signals live in the same *(C x N)*-dimensional vector space, for some *N* large enough.
The delay offset is merely an implementation detail to avoid lots of zeros.

Compare to the [sparse matrix classes](https://docs.scipy.org/doc/scipy-0.18.1/reference/generated/scipy.sparse.coo_matrix.html) which (of course) do matrix arithmetics.

mgeier:
OK, we might need signal superposition at some point, but is it something that has influence on the current PR?

I think we can decide later if we use operator overloading or not.

But the decision if we use integer or fractional delays indeed has a lot of influence here.
As far as I understood, we want fractional delays, right?

The relation to sparse matrices is very interesting, but as you say, it's only valid for integer delays.

And even then, there is the difference that we need a way to align the signal data on the time axis.

Another detail is that our signals (currently) cannot have "holes". If you add a signal that is non-zero in the first second to another signal that only plays in the 10th second, you'll have to fill the time in-between with zeros.

> The delay offset is merely an implementation detail to avoid lots of zeros.

No, it's not.
It would be if we only used positive delays, but it is actually required for negative delays. Or we would need to set some arbitrary maximum negative delay somewhere.
Or do you know a different way how we could implement this?

Another possibility would be to use one delay value per channel instead of one global delay for all channels.
Instead of `(data1, fs, -0.007)`, we could have something like `(data2, fs, [0.02, -0.005, 0.006, 0.001, -0.007])` (assuming 5 channels).
We could even make it optional: A scalar value would be applied to all channels, a sequence would have to have the right length and its element would be applied column-wise.

What about that?

But then we also might think about having numeric weights per channel instead of applying the weights destructively ...

trettberg:
> OK, we might need signal superposition at some point, but is it something that has influence on the current PR?

Go ahead!

> The relation to sparse matrices is very interesting, but as you say, it's only valid for integer delays.

Mathematically it's the same fractional delays (even with multiple sampling rates).
All our signals live in the same vector space.
(Implementation is a different thing though, that's true.)

I think we should stick with a simple `DelayedSignal` class, that is a single delay per signal:
* The data part can always be used "raw" with consistent time alignment.
Nobody would be forced to use the custom functionality.
* For the same reason, it's probably a good thing that signals cannot have "holes".
* When using multiple signals, manual time alignment is IMHO much more tedious than manual weighting. So I don't see why `DelayedSignal` should care about individual channel weights.

spors:
We should go ahead with the class and check its application in the toolbox.

mgeier:
OK, I've merged it, thanks for all the comments!

Of course this is not set in stone, we can still change any aspect of it ...

Issue #28, opened 2016-12-11
Naming: Clarify what "normalization" means


mgeier:
Currently we have `sfs.util.normalize()` and `sfs.util.normal_vector()` which do very different things.
This might be confusing.

spors:
The function  `sfs.util.normal_vector()` takes a vector and normalizes it. It could therefore rename it to  `sfs.util.normalize_vector()`.

trettberg:
Alternatively we could get rid of `sfs.util.normal_vector()`. It seems it is used only in `sfs.util.rotation_matrix()`.

mgeier:
@trettberg That may be its only use for now, but we should call it wherever normal vectors are passed to functions, see #20.

chris-hld:
I just stumbled about this problem.
I'll need a `sfs.util.normalize_vector()`.
Please do not remove ;)

hagenw:
In #34 I introduced now `util.normalize_vector()`. I'm still not completely sure if this is the best name. Most of the time we are dealing with directions when we are using unit vectors. We have also `util.direction_vector(alpha, beta=np.pi/2)` and I could also imagine `util.direction_vector(point1, point2)`. Maybe there is a way to combine all this?

chris-hld:
`util.direction_vector(point1, point2)` would be very handy, if not necessary.
Or I would introduce an optional "base" to `normalize_vector()` . The default would be 0.

Please have a look at this:
>xs = np.r_[0.5, 0.5, 0]  # position of virtual source
>xref = np.r_[0, 0, 0]
>nfs = sfs.util.normalize_vector(xref - xs)

For subtracting, I first have to create an array, which is not consistent with our array_like structure elsewhere. But actually I just have to specify the base of the vector to be `xs`.

Should we rather introduce a new function, or extend the existing?

mgeier:
I agree that we need a way to subtract vectors (without having a vector data type), but I'm not sure if it is obvious enough what a function named `direction_vector()` with two points as input would do.

Does it just subtract or also normalize?

Also, Python doesn't do function overloading, so having the same `direction_vector()` function with alpha/beta and point1/point2 is a bit awkward.

The concept of a "base" in this context is new to me, I'm not sure if that's easy enough to understand.

What we need is something similar to a "look at" function.
Just some random links:
https://docs.unity3d.com/ScriptReference/Transform.LookAt.html
[...]

PR #34, opened 2016-12-21
Fix non-unit direction vectors

### LINKED ISSUE #20: Handling of directional vector for plane waves
At the moment we can change the frequency of the resulting sound field of a plane wave by using non unity vectors for specifying the direction. I don't know if this is desired behavior, but I would suggest to change it to at least raise a warning or an error. Or maybe to some automatic normalization of the directional vector.

```python
import numpy as np
import sfs

grid = sfs.util.xyz_grid([-2, 2], [-2, 2], 0, spacing=0.01)
x0 = 0, 0, 0
npw = 0, -1, 0
f = 1000
omega = 2 * np.pi * f
p_plane = sfs.mono.source.plane(omega, x0, npw, grid);
sfs.plot.soundfield(p_plane, grid);
plt.title('Plane wave with npw = {}'.format(npw));
plt.savefig('pw1.png')

npw = 0, -2, 0
p_plane = sfs.mono.source.plane(omega, x0, npw, grid);
sfs.plot.soundfield(p_plane, grid);
plt.title('Plane wave with npw = {}'.format(npw));
plt.savefig('pw2.png')
```

pw1

pw2

hagenw:
This tries to fix #20 and #28.

I renamed `util.normal_vector()` to `util.normalize_vector()` as proposed in #28 and used it for every single dimensional direction vector `n`. The field plotted in #20 looks now correct with this.

I'm still not happy with the naming, which I will discuss directly in #28 and there is the problem how to handle several unit vectors at once like `n0 = util.asarray_of_rows(n0)`.

### DIFF
--- a/sfs/mono/drivingfunction.py
+++ b/sfs/mono/drivingfunction.py
@@ -86,7 +86,7 @@ def _wfs_plane(omega, x0, n0, n=[0, 1, 0], c=None):
     """
     x0 = util.asarray_of_rows(x0)
     n0 = util.asarray_of_rows(n0)
-    n = util.asarray_1d(n)
+    n = util.normalize_vector(n)
     k = util.wavenumber(omega, c)
     return 2j * k * np.inner(n, n0) * np.exp(-1j * k * np.inner(n, x0))
 
@@ -106,7 +106,7 @@ def wfs_25d_plane(omega, x0, n0, n=[0, 1, 0], xref=[0, 0, 0], c=None,
     """
     x0 = util.asarray_of_rows(x0)
     n0 = util.asarray_of_rows(n0)
-    n = util.asarray_1d(n)
+    n = util.normalize_vector(n)
     xref = util.asarray_1d(xref)
     k = util.wavenumber(omega, c)
     return wfs_25d_preeq(omega, omalias, c) * \
@@ -179,7 +179,7 @@ def wfs_25d_preeq(omega, omalias, c):
 def delay_3d_plane(omega, x0, n0, n=[0, 1, 0], c=None):
     """Plane wave by simple delay of secondary sources."""
     x0 = util.asarray_of_rows(x0)
-    n = util.asarray_1d(n)
+    n = util.normalize_vector(n)
     k = util.wavenumber(omega, c)
     return np.exp(-1j * k * np.inner(n, x0))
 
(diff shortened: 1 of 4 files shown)


mgeier:
Any objections?
If not, LGTM!

hagenw:
Just a reminder, the fix is not done for `n0 = util.asarray_of_rows(n0)` as I didn't know what is the best way to apply it there. With a loop? Can someone help out with this?

mgeier:
If this would be needed only once, you could do something like:

```python
n0 = util.asarray_of_rows(n0)
n0 = np.row_stack([util.normalize_vector(n) for n in n0])
```

But since we need this several times, we should make a function.

Should we include this into `util.normalize_vector()`?
We could do a simple `np.asarray()` and then do different things for one and two-dimensional arrays.
This could be used like `util.normalize_vector(util.asarray_of_rows(n0))`

Or should we make a new function `util.asarray_of_normalized_rows()`?
This has the problem of #28, could be both types of "normalization".

In a separate function, you could use something like this:

```python
# make sure x is a 2-dimensional array
x / np.linalg.norm(x, axis=1, keepdims=True)
```

However, note that the `keepdims` argument is only available since NumPy 1.10.0 (see https://docs.scipy.org/doc/numpy/reference/generated/numpy.linalg.norm.html).

This should do the same thing and also work for older versions:

```python
# make sure x is a 2-dimensional array
x / np.linalg.norm(x, axis=1).reshape(-1, 1)
```

mgeier:
PS: if we include this in `util.normalize_vector()`, it will lose its squeezing ability.
This means we would have to use `util.normalize_vector(util.asarray_1d(n))` in many cases.

Which would be a bit longer, but it would actually be more explicit, which generally is a good thing.

Another problem: we would have to decide between `normalize_vector()` and `normalize_vectors()`.

BTW, if we decide for some fundamental changes along the lines of #36, we'll have to change those functions again.

mgeier:
PPS: We can also implement this later, we should just create a new issue as a reminder.

hagenw:
At the beginning I was also thinking of two functions similar to the `asarray` once, e.g.
```python
util.asdirection()
util.asdirections()
```
But maybe it is better to have to use it in an explicit way?
```python
util.normalize_vector(util.asarray_1d(n))
util.normalize_vector(util.asarray_of_rows(n))
```

mgeier:
I would suggest to merge this PR as is and postpone further changes until we see how our understanding of a "direction" evolves.

hagenw:
Fine with me, I created an issue for the remaining TODO of this PR.

PR #35, opened 2016-12-21
Add time based 2.5D-WFS focused source 

chris-hld:
This implements the simulation of a virtual focused sound source for time based 2.5D-WFS.
 - The documentation is updated accordingly.
 - There is also an example.

### DIFF
--- a/examples/time_domain.py
+++ b/examples/time_domain.py
@@ -62,4 +62,27 @@ plt.grid()
 sfs.plot.virtualsource_2d([0, 0], npw, type='plane')
 plt.title('impulse_pw_wfs_25d')
 plt.savefig('impulse_pw_wfs_25d.png')
+
+# FOCUSED SOURCE
+xs = np.r_[0.5, 0.5, 0]  # position of virtual source
+xref = np.r_[0, 0, 0]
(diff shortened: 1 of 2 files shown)


examples/time_domain.py
```diff
@@ -63,4 +63,28 @@
 sfs.plot.virtualsource_2d([0, 0], npw, type='plane')
 plt.title('impulse_pw_wfs_25d')
 plt.savefig('impulse_pw_wfs_25d.png')
+
+# FOCUSED SOURCE
+xs = np.r_[0.5, 0.5, 0]  # position of virtual source
+xref = np.r_[0, 0, 0]
+nfs = sfs.util.normal_vector(xref - xs)  # main n of fsource
```
chris-hld: This needs review. See #28

mgeier:
Thanks!

I think the problem you mention in https://github.com/sfstoolbox/sfs-python/pull/35#discussion_r93488651 will take some more time to resolve ...

I suggest merging this as-is right now and fix the problem with subtracting coordinates later.

Any objections?

chris-hld:
yep, then the basic time domain 2.5D WFS driving functions are implemented (except the filter).

mgeier:
I've taken the liberty of squashing and rebasing your commits. Thanks for this PR!

PR #38, opened 2017-01-24
Visualization of (continuous) secondary source contour by polygon

spors:
```
import numpy as np
import sfs

xs = [0, 2, 0]
grid = sfs.util.xyz_grid([-2.5, 2.5], [-2.5, 2.5], 0, spacing=0.02)
omega = 2*np.pi*1000

x0, n0, a0 = sfs.array.circular(200, 1.5)
d = sfs.mono.drivingfunction.wfs_2d_line(omega, x0, n0, xs)
a = sfs.mono.drivingfunction.source_selection_point(n0, x0, xs)
twin = sfs.tapering.tukey(a, 0.3)
p = sfs.mono.synthesized.generic(omega, x0, n0, d * a0 * twin, grid)

sfs.plot.soundfield(p, grid, xnorm=[0, 0, 0], vmin=-1.5, vmax=1.5)
sfs.plot.secondarysourcecontour_2d(x0, a)
sfs.plot.virtualsource_2d(xs)
```

grafik

### DIFF
--- a/sfs/plot.py
+++ b/sfs/plot.py
@@ -3,7 +3,8 @@ from __future__ import division
 import matplotlib.pyplot as plt
 from matplotlib.patches import PathPatch
 from matplotlib.path import Path
-from matplotlib.collections import PatchCollection
+from matplotlib.collections import PatchCollection, LineCollection
+from matplotlib.colors import ListedColormap
 from mpl_toolkits import axes_grid1
 from mpl_toolkits.mplot3d import Axes3D
 import numpy as np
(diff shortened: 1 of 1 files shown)


trettberg:
Two observations:
* The contour has a gap.
* The region of active sources is not symmetric. (Easier to see with few speakers, 16 or so).

mgeier:
I guess this needs a `closed` parameter like `sfs.array.weights_midpoint()`?

mgeier:
Possible fix for the second point made by @trettberg above: instead of line segments from one secondary source to the next, we should use line segments from halfway from the previous source to the current source to halfway through the next source. Again, the implementation could be similar to `sfs.array.weights_midpoint()`.

spors:
I have fixed the issue regarding the gap and the region of active secondary sources:
```
import numpy as np
import sfs

xs = [5, 0, 0]

x0, n0, a0 = sfs.array.circular(20, 1.5)
a = sfs.mono.drivingfunction.source_selection_point(n0, x0, xs)

plt.figure(figsize=(10, 10))
sfs.plot.secondarysourcecontour_2d(x0, a)
sfs.plot.loudspeaker_2d(x0, n0, a, size=0.2)
sfs.plot.virtualsource_2d(xs)
plt.axis([-5.2, 5.2, -5.2, 5.2])
```
grafik

sfs/plot.py
```diff
@@ -86,6 +87,43 @@ def secondarysource_2d(x0, n0, grid=None):
...
+
+    # set color of line to indicate active parts
+    a0 = np.append(a0, a0[0])
+    active = np.logical_and(a0[0:-1], a0[1:])
+    lc.set_array(active)
+
+    # add collection of lines to current axis
+    if ax is None:
+        ax = plt.gca()
+    plt.gca().add_collection(lc)
```
mgeier: This should be `ax.add_collection(lc)`

mgeier:
Now it looks like the array is always closed, even if we don't want it to be:

```python
sfs.plot.secondarysourcecontour_2d([(0, 0), (1, 1), (2, 1), (3, 0)],
                                   [False, True, True, False])
plt.axis('equal');
```

image

hagenw:
Have the secondary sources to be connected by a line?
Under Matlab I solved the gap issue by using only points for plotting secondary sources. If the distance between them becomes very small then they appear as a line. If the distance is not small, points have the advantage that you still are able to see the actual distance.

PR #41, opened 2017-09-01
DOC: Change documentation of "signal" arguments

mgeier:
... discouraging the use of an explicit delay value.


mgeier:
@hagenw I added "Hertz" and "seconds" in 2c769d00f503d9f8a253d01578d83b0b200b8d71

PR #44, opened 2017-11-07
Switch to spherical_jn and spherical_yn

### LINKED ISSUE #43: Not compatible with scipy >=0.19
`sph_jnyn` which we [import in `drivingfunction.py`](https://github.com/sfstoolbox/sfs-python/blob/master/sfs/mono/drivingfunction.py#L9) was set to be [deprecated in scipy 0.18.0](https://docs.scipy.org/doc/scipy/reference/release.0.18.0.html#deprecated-features) and is no longer availabe in scipy 0.19.0.

I guess we have to switch to [`spherical_jn`](https://docs.scipy.org/doc/scipy-0.19.1/reference/generated/scipy.special.spherical_jn.html#scipy.special.spherical_jn) and [`spherical_yn`](https://docs.scipy.org/doc/scipy-0.19.1/reference/generated/scipy.special.spherical_jn.html#scipy.special.spherical_yn).

hagenw:
This tries to fix #43.

Unfortunately, the return values of the spherical Bessel functions seem to have changed as well, even if the documentation still states `ndarray`.
If you execute the following code with my changes you will get an error:
```python
import numpy as np
import sfs
nk = 0, -1, 0  # direction of plane wave
omega = 2 * np.pi * 1000  # frequency
R0 = 1.5  # radius of secondary sources
x0, n0, a0 = sfs.array.circular(200, R0)
grid = sfs.util.xyz_grid([-1.75, 1.75], [-1.75, 1.75], 0, spacing=0.02)
d = sfs.mono.drivingfunction.nfchoa_25d_plane(omega, x0, R0, nk)
```
The error is
```
IndexError                                Traceback (most recent call last)
<ipython-input-5-5d956e53e92e> in <module>()
----> 1 d = sfs.mono.drivingfunction.nfchoa_25d_plane(omega, x0, R0, nk)

/scratch/git/sfs-python/sfs/mono/drivingfunction.py in nfchoa_25d_plane(omega, x0, r0, n, max_order, c)
    313     d = 0
    314     for m in range(-M, M + 1):
--> 315         d += 1j**-abs(m) / (k * hn2[abs(m)]) * np.exp(1j * m * (phi0 - phi))
    316     return -2 / r0 * d
    317 

IndexError: invalid index to scalar variable.
```

### DIFF
--- a/sfs/util.py
+++ b/sfs/util.py
@@ -3,6 +3,7 @@
 from __future__ import division
 import collections
 import numpy as np
+from scipy.special import spherical_jn, spherical_yn
 from . import defs
 
 
@@ -491,3 +492,27 @@ def image_sources_for_box(x, L, N, prune=True):
         wall_count = wall_count[N_mask, :]
 
     return xs, wall_count
+
+
+def spherical_hn2(n, z):
+    r"""Spherical Hankel function of 2nd kind.
+
+    Defined as http://dlmf.nist.gov/10.47.E6,
+
+    .. math::
+
+        \hankel{2}{n}{z} = \sqrt{\frac{\pi}{2z}}
+        \Hankel{2}{n + \frac{1}{2}}{z},
+
+    where :math:`\Hankel{2}{n}{\cdot}` is the Hankel function of the
+    second kind and n-th order, and :math:`z` its complex argument.
+
+    Parameters
+    ----------
+    n : array_like
+        Order of the spherical Hankel function (n >= 0).
+    z : array_like
+        Argument of the spherical Hankel function.
+
+    """
+    return spherical_jn(n, z) - 1j * spherical_yn(n, z)
--- a/sfs/util.py
+++ b/sfs/util.py
@@ -1,4 +1,8 @@
-"""Various utility functions."""
+"""Various utility functions.
+
+.. include:: math-definitions.rst
+
+"""
 
 from __future__ import division
 import collections
--- a/sfs/mono/drivingfunction.py
+++ b/sfs/mono/drivingfunction.py
@@ -6,7 +6,7 @@
 
 import numpy as np
 from numpy.core.umath_tests import inner1d  # element-wise inner product
-from scipy.special import sph_jnyn, jn, hankel2
+from scipy.special import jn, spherical_jn, spherical_yn, hankel2
 from .. import util
 from .. import defs
 
@@ -281,8 +281,8 @@ def nfchoa_25d_point(omega, x0, r0, xs, max_order=None, c=None):
     phi, _, r = util.cart2sph(*xs)
     phi0 = util.cart2sph(*x0.T)[0]
     M = _max_order_circular_harmonics(len(x0), max_order)
-    hr = _sph_hn2(M, k * r)
-    hr0 = _sph_hn2(M, k * r0)
+    hr = _spherical_hn2(range(0, M + 1), k * r)
+    hr0 = _spherical_hn2(range(0, M + 1), k * r0)
     d = 0
     for m in range(-M, M + 1):
         d += hr[abs(m)] / hr0[abs(m)] * np.exp(1j * m * (phi0 - phi))
@@ -309,8 +309,8 @@ def nfchoa_25d_plane(omega, x0, r0, n=[0, 1, 0], max_order=None, c=None):
     phi, _, r = util.cart2sph(*n)
     phi0 = util.cart2sph(*x0.T)[0]
     M = _max_order_circular_harmonics(len(x0), max_order)
-    hn2 = _sph_hn2(M, k * r0)
     d = 0
+    hn2 = _spherical_hn2(range(0, M + 1), k * r0)
     for m in range(-M, M + 1):
         d += 1j**-abs(m) / (k * hn2[abs(m)]) * np.exp(1j * m * (phi0 - phi))
     return -2 / r0 * d
@@ -669,10 +669,27 @@ def esa_edge_dipole_2d_line(omega, x0, xs, alpha=3/2*np.pi, Nc=None, c=None):
     return -1j*np.pi/alpha * d
 
 
-def _sph_hn2(n, z):
-    """Spherical Hankel function of 2nd kind."""
-    jn, jnp, yn, ynp = sph_jnyn(n, z)
-    return jn - 1j * yn
+def _spherical_hn2(n, z):
+    r"""Spherical Hankel function of 2nd kind.
+
+    Defined as http://dlmf.nist.gov/10.47.E6,
+
+    .. math:: \hankel{2}{n}{z} = \sqrt{\frac{\pi}{2z}} \Hankel{2}{n + 1/2}{z},
+
+    where :math:`\Hankel{2}{n}{}` is the Hankel function of the second kind.
+
+    Parameters
+    ----------
+    n : int, array_like
+        Order of the spherical Hankel function (n >= 0).
+    z : complex or float, array_like
+        Argument of the spherical Hankel function.
+
+    Returns
+    -------
+    hn2 : ndarray
+    """
+    return spherical_jn(n, z) - 1j * spherical_yn(n, z)
 
 
 def _max_order_circular_harmonics(N, max_order):
--- a/sfs/mono/drivingfunction.py
+++ b/sfs/mono/drivingfunction.py
@@ -6,7 +6,7 @@
 
 import numpy as np
 from numpy.core.umath_tests import inner1d  # element-wise inner product
-from scipy.special import jn, spherical_jn, spherical_yn, hankel2
+from scipy.special import jn, hankel2
 from .. import util
 from .. import defs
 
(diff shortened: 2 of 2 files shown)


hagenw:
Indeed, they no longer return an array.

sfs 0.3.1 and scipy 0.17.0:
```python
In: sfs.mono.drivingfunction._sph_hn2(3, 1)
Out: 
array([[ 0.84147098 +0.54030231j,  0.30116868 +1.38177329j,
         0.06203505 +3.60501757j,  0.00900658+16.64331454j],
       [-0.30116868 -1.38177329j,  0.23913363 -2.22324428j,
         0.11506352 -9.43327941j,  0.02600873-62.96824059j]])
```

This pull request and scipy 1.0.0:
```python
In: sfs.mono.drivingfunction._spherical_hn2(3, 1)
Out: (0.0090065811171125242+16.643314540123811j)
```

Would you have to do instead to get the correct result is
```python
In: sfs.mono.drivingfunction._spherical_hn2(range(3+1), 1)
Out: 
array([ 0.84147098 +0.54030231j,  0.30116868 +1.38177329j,
        0.06203505 +3.60501757j,  0.00900658+16.64331454j])
```

hagenw:
My question then:
Should we adjust `_spherical_hn2` to behave the same and return only one value if one order was given, or should we continue with the same behaviour as before and always return an array by adjusting the function to:
```python
def _spherical_hn2(n, z):
    """Spherical Hankel function of 2nd kind."""
    jn = spherical_jn(range(n+1), z)
    yn = spherical_yn(range(n+1), z)
    return jn - 1j * yn
```

mgeier:
It doesn't make sense to return multiple values from our helper function when we use `spherical_jn()` and `spherical_yn()`.

We did this previously because `sph_jnyn()` did it, presumably for performance reasons.

Now, we should just return a single complex value and call the helper function repeatedly in the loop.

hagenw:
I updated it accordingly.

I also added a doc-string to `_spherical_hn2` similar to the one from `spherical_jn`, but as it is only an internal function this is of course not included in the online documentation. Do you would prefer then to not have such a long doc-string or what is the general rule in this case?

spors:
Most of the functions for spherical Bessel/Hankel function returned orders up to a given order N because the reoccurence relations of the spherical Bessel/Hankel function allow an efficient calculation. If we calculate every order seperately, this could increase the computational complexity considerably.

hagenw:
`_spherical_hn2` has the same behaviour as `spherical_jn` and can return the calculation for multiple orders.
I could change the places where it is used back to the old behaviour, or we do some performance tests in order to see if this is necessary.

mgeier:
> can return the calculation for multiple orders

Yes, but that doesn't necessarily mean that it is more efficient than multiple calls (not considering the function call overhead).

> do some performance tests

Yes. We should only calculate multiple orders at once if it improves performance.

hagenw:
>> can return the calculation for multiple orders
>
>Yes, but that doesn't necessarily mean that it is more efficient than multiple calls (not considering the function call overhead).

It's very likely as the calculation is recurrent, always starting at the lowest order independent of the asked one.

hagenw:
I did a short performance test by running the following code.
```python
import timeit
from scipy.special import spherical_jn, spherical_yn

def _spherical_hn2(n, z):
    return spherical_jn(n, z) - 1j * spherical_yn(n, z)

z = 20
max_order = 150

start_time = timeit.default_timer()
for n in range(0, max_order):
    _spherical_hn2(n, z)
elapsed = timeit.default_timer() - start_time
print(elapsed)

start_time = timeit.default_timer()
a = _spherical_hn2(range(0, max_order), z)
for n in range(0, max_order):
    a[n]
elapsed = timeit.default_timer() - start_time
print(elapsed)
```
Output:
```
0.0027250559942331165
0.0006347029993776232
```
As expected the performance is better for asking directly for several orders as they only have to be calculated once. On the other hand this is already the worst case scenario as orders higher than 150 will result in numerical problems anyhow and will not be used in practice.

As the results were faster with the calculation outside of the loop, I changed the code accordingly to its old behavior.

mgeier:
The time difference might be solely caused by the overhead of calling into some C functions.
I still don't think that any recurrence relations are used.

I think it would be more interesting to add the rest of the math operations that are done in the loop to the benchmark.

Only then, if there is still a significant difference, should we do the calculation at once.

BTW, using `sph_jnyn()` is another 10x faster: https://gist.github.com/mgeier/51d8758eb94e8f6acaa1249494dfd16e

mgeier:
I've updated my Gist: https://gist.github.com/mgeier/51d8758eb94e8f6acaa1249494dfd16e

Please check if the benchmarks make sense.

I added some operations inside the loop.
This doesn't change much in comparison to before, calling `_spherical_hn2()` outside of the loop is a 6x improvement on my computer.
The interesting thing, however, is that switching to `sph_jnyn()` only gives less than 2x speed!

If my observations are correct we should merge this PR!

> I also added a doc-string to `_spherical_hn2` similar to the one from `spherical_jn`, but as it is only an internal function this is of course not included in the online documentation. Do you would prefer then to not have such a long doc-string or what is the general rule in this case?

AFAIK there is no well-established rule about this.
IMO there is nothing wrong with detailed docstrings on internal functions.
Having a lot of reST markup in them doesn't make too much sense though, since it is never actually rendered anywhere.

Since you made this nice docstring, we should probably turn the function into an "official" function, something like `sfs.util.spherical_hn2()`?

sfs/util.py
```diff
@@ -491,3 +496,26 @@ def _count_walls_1d(a):
...
 
     return xs, wall_count
+
+
+def spherical_hn2(n, z):
+    r"""Spherical Hankel function of 2nd kind.
+
+    Defined as http://dlmf.nist.gov/10.47.E6,
+
+    .. math:: \hankel{2}{n}{z} = \sqrt{\frac{\pi}{2z}} \Hankel{2}{n + 1/2}{z},
```
mgeier: I think `\frac{1}{2}` would be clearer, because otherwise one might be tempted to think at first glance it's `(n+1)/2`.

sfs/util.py
```diff
@@ -491,3 +496,26 @@ def _count_walls_1d(a):
...
+
+    where :math:`\Hankel{2}{n}{}` is the Hankel function of the second kind.
```
mgeier: I'd prefer a `\cdot` in the parentheses, but how is this handled in the SFS theory?

sfs/util.py
```diff
@@ -491,3 +496,26 @@ def _count_walls_1d(a):
...
+
+    Parameters
+    ----------
+    n : int, array_like
```
mgeier: IIRC, giving the datatype here is not what we normally do in the SFS toolbox.

We would normally just use `array_like` or we prepend the dimensions.

I know this is how SciPy does it, but shall we do it, too?

In NumPy I didn't find this style, I only found e.g. "int or 1-D array_like".

sfs/util.py
```diff
@@ -491,3 +496,26 @@ def _count_walls_1d(a):
...
+        Order of the spherical Hankel function (n >= 0).
+    z : complex or float, array_like
+        Argument of the spherical Hankel function.
+
+    Returns
+    -------
+    hn2 : ndarray
```
mgeier: This style is inconsistent with the rest of the file and the toolbox.

Also note that there should be an empty line before the closing `"""`.

sfs/util.py
hagenw: In most cases in SFS theory I wrote it without any argument, e.g. `f(x) = g(x) + h(x), where `g` is a function. For the Hankel function this is not possible as we have to provide a third argument. In this case I agree with you that adding a `\cdot` should be better.

sfs/util.py
hagenw: The input can be a single int or a array_like thing including only ints.
For cases in which the input is a single value we provide `int` or `float` as description.
Regarding the dimensions, I guess they don't matter, but I'm not sure I just copied the description from SciPy.

By the way in the Toolbox I found also this one: [`int(N, 3) array_like`](https://github.com/sfstoolbox/sfs-python/blob/master/sfs/mono/drivingfunction.py#L404).

mgeier: Well I guess `int(N, 3) array_like` is a mistake (that was copy-pasted a few times), it doesn't make any sense at all.

According to the NumPy style we could write "int or array_like", but this wouldn't really make sense either, because an "int" is already a valid "array_like" (it gets converted to a zero-dimensional array).
We could write "array_like of int" or "array_like of ints" but I think this doesn't look quite nice.
We could write something like "array_like, dtype=int", which would be similar to how scikit learn does it, but I think that's more complicated than necessary.

I think it would be best to just write "array_like". It should be clear that the "order" has to be an integer.

hagenw: OK, I switched to `array_like`.

sfs/util.py
```diff
@@ -491,3 +496,31 @@ def _count_walls_1d(a):
...
+    Parameters
+    ----------
+    n : array_like
+        Order of the spherical Hankel function (n >= 0).
+    z : complex or float, array_like
+        Argument of the spherical Hankel function.
+
+    Returns
+    -------
+    hn2 : array_like
```
mgeier: I wouldn't invent a variable name for a single return value.
I think this only helps if there are multiple return values.

I normally don't use "array_like" for return values, because those are typically more concrete types, like "numpy.ndarray".

However, in this case there are multiple possibilities ... to make this clear, we could write something like

```rst
    Returns
    -------
    complex scalar or numpy.ndarray
        The function value(s) for the given order(s) and argument(s).
```

But this seems unnecessarily verbose, doesn't it?

I think in this case we could also completely drop the whole "Returns" section, because neither the return type nor the description is particularly interesting or surprising.
The only real information in there is that the function values are complex. But if we need to say that, we can also just mention it in the main description.

sfs/util.py
```diff
@@ -491,3 +496,31 @@ def _count_walls_1d(a):
...
+        \Hankel{2}{n + \frac{1}{2}}{z},
+
+    where :math:`\Hankel{2}{n}{\cdot}` is the Hankel function of the
+    second kind.
+
+    Parameters
+    ----------
+    n : array_like
+        Order of the spherical Hankel function (n >= 0).
+    z : complex or float, array_like
```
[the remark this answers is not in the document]
hagenw: I adjusted it to `array_like` and descriped `z` as a complex argument in the description.

sfs/util.py
hagenw: I found it also verbose, that was the reason why I skipped the description of the return value.
I removed it now completely.

PR #45, opened 2017-11-08
NFC-HOA second-order-section filters

narahahn:
NFC-HOA driving functions are implemented using first/second-order section filters.

* based on [Spors et al](http://www.int.uni-rostock.de/fileadmin/user_upload/publications/spors/2011/Spors_etal_WASPAA_2011_efficient_NFCHOA.pdf) and [Hahn and Spors](http://www.aes.org/e-lib/browse.cfm?elib=18609)
* driving functions for plane wave and point source

ps_level    ps_soundfield
pw_level pw_soundfield

To be discussed
* [x] The SOS filter coefficients are stored in a dictionary (see below). Is there a more elegant way?
```
sos = {}
for m in range(0, max_order+1):
     ...
     sos[m] = sig.zpk2sos(z0, zinf, k, pairing='nearest')
```
* [ ] I realized that the Ambisonic order can be increased beyond 150 by doing some hack on `scipy.signal.besselap`. But I'm not so convinced whether this is really necessary, considering that the cut-off frequency for 150th order is already around 20 kHz.
* [x] So far, the poles and zeros in the Laplace domain is converted into the z-domain by using the matched-z transform. If necessary, bilinear transform can be added.

### DIFF
--- a/doc/examples/time_domain_nfchoa.py
+++ b/doc/examples/time_domain_nfchoa.py
@@ -1,0 +1,50 @@
+"""Create some examples of time-domain NFC-HOA."""
+
+import numpy as np
+import matplotlib.pyplot as plt
+import sfs
+from scipy.signal import unit_impulse
+
+# Parameters
+fs = 44100  # sampling frequency
+grid = sfs.util.xyz_grid([-2, 2], [-2, 2], 0, spacing=0.005)
+N = 60  # number of secondary sources
+R = 1.5  # radius of circular array
+array = sfs.array.circular(N, R)
+
+# Excitation signal
+signal = unit_impulse(512), fs, 0
+
+# Plane wave
+max_order = None
+npw = [0, -1, 0]  # propagating direction
+t = 0  # observation time
+delay, weight, sos, phaseshift, selection, secondary_source = \
+    sfs.time.drivingfunction.nfchoa_25d_plane(array.x, R, npw, fs, max_order)
+d = sfs.time.drivingfunction.nfchoa_25d_driving_signals(
+        delay, weight, sos, phaseshift, signal)
+p = sfs.time.synthesize(d, selection, array, secondary_source,
+                        observation_time=t, grid=grid)
+
+plt.figure()
+sfs.plot.level(p, grid)
+sfs.plot.loudspeaker_2d(array.x, array.n)
+sfs.plot.virtualsource_2d([0, 0], ns=npw, type='plane')
+plt.savefig('impulse_pw_nfchoa_25d.png')
+
+# Point source
+max_order = 100
+xs = [1.5, 1.5, 0]  # position
+t = np.linalg.norm(xs) / sfs.default.c  # observation time
+delay, weight, sos, phaseshift, selection, secondary_source = \
+    sfs.time.drivingfunction.nfchoa_25d_point(array.x, R, xs, fs, max_order)
+d = sfs.time.drivingfunction.nfchoa_25d_driving_signals(
+        delay, weight, sos, phaseshift, signal)
+p = sfs.time.synthesize(d, selection, array, secondary_source,
+                        observation_time=t, grid=grid)
+
+plt.figure()
+sfs.plot.level(p, grid)
+sfs.plot.loudspeaker_2d(array.x, array.n)
+sfs.plot.virtualsource_2d(xs, type='point')
+plt.savefig('impulse_ps_nfchoa_25d.png')
--- a/sfs/time/drivingfunction.py
+++ b/sfs/time/drivingfunction.py
@@ -7,8 +7,8 @@
 
     import matplotlib.pyplot as plt
     import numpy as np
-    from scipy.signal import unit_impulse
     import sfs
+    from scipy.signal import unit_impulse
 
     # Plane wave
     npw = sfs.util.direction_vector(np.radians(-45))
@@ -25,7 +25,8 @@
     tf = rf / sfs.default.c  # time-of-arrival at origin
 
     # Impulsive excitation
-    signal = unit_impulse(512), 44100
+    fs = 44100
+    signal = unit_impulse(512), fs
 
     # Circular loudspeaker array
     N = 32  # number of loudspeakers
@@ -43,6 +44,8 @@
 """
 import numpy as np
 from numpy.core.umath_tests import inner1d  # element-wise inner product
+from scipy.signal import besselap, sosfilt, zpk2sos
+from scipy.special import eval_legendre as legendre
 from .. import default
 from .. import util
 from . import source as _source
@@ -357,3 +360,448 @@ def secondary_source_point(c):
         return _source.point(position, signal, observation_time, grid, c=c)
 
     return secondary_source
+
+
+def matchedz_zpk(s_zeros, s_poles, s_gain, fs):
+    """Matched-z transform of poles and zeros.
+
+    Parameters
+    ----------
+    s_zeros : array_like
+        Zeros in the Laplace domain.
+    s_poles : array_like
+        Poles in the Laplace domain.
+    s_gain : float
+        System gain in the Laplace domain.
+    fs : int
+        Sampling frequency in Hertz.
+
+    Returns
+    -------
+    z_zeros : numpy.ndarray
+        Zeros in the z-domain.
+    z_poles : numpy.ndarray
+        Poles in the z-domain.
+    z_gain : float
+        System gain in the z-domain.
+
+    See Also
+    --------
+    :func:`scipy.signal.bilinear_zpk`
+
+    """
+    z_zeros = np.exp(s_zeros / fs)
+    z_poles = np.exp(s_poles / fs)
+    omega = 1j * np.pi * fs
+    s_gain *= np.prod((omega - s_zeros) / (omega - s_poles)
+                      * (-1 - z_poles) / (-1 - z_zeros))
+    return z_zeros, z_poles, np.real(s_gain)
+
+
+def nfchoa_25d_plane(x0, r0, npw, fs, max_order=None, c=None,
+                     s2z=matchedz_zpk):
+    r"""Virtual plane wave by 2.5-dimensional NFC-HOA.
+
+    .. math::
+
+        D(\phi_0, s) =
+        2\e{\frac{s}{c}r_0}
+        \sum_{m=-M}^{M}
+        (-1)^m
+        \Big(\frac{s}{s-\frac{c}{r_0}\sigma_0}\Big)^\mu
+        \prod_{l=1}^{\nu}
+        \frac{s^2}{(s-\frac{c}{r_0}\sigma_l)^2+(\frac{c}{r_0}\omega_l)^2}
+        \e{\i m(\phi_0 - \phi_\text{pw})}
+
+    The driving function is represented in the Laplace domain,
+    from which the recursive filters are designed.
+    :math:`\sigma_l + \i\omega_l` denotes the complex roots of
+    the reverse Bessel polynomial.
+    The number of second-order sections is
+    :math:`\nu = \big\lfloor\tfrac{|m|}{2}\big\rfloor`,
+    whereas the number of first-order section :math:`\mu` is either 0 or 1
+    for even and odd :math:`|m|`, respectively.
+
+    Parameters
+    ----------
+    x0 : (N, 3) array_like
+        Sequence of secondary source positions.
+    r0 : float
+        Radius of the circular secondary source distribution.
+    npw : (3,) array_like
+        Unit vector (propagation direction) of plane wave.
+    fs : int
+        Sampling frequency in Hertz.
+    max_order : int, optional
+        Ambisonics order.
+    c : float, optional
+        Speed of sound in m/s.
+    s2z : callable, optional
+        Function transforming s-domain poles and zeros into z-domain,
+        e.g. :func:`matchedz_zpk`, :func:`scipy.signal.bilinear_zpk`.
+
+    Returns
+    -------
+    delay : float
+        Overall delay in seconds.
+    weight : float
+        Overall weight.
+    sos : list of numpy.ndarray
+        Second-order section filters :func:`scipy.signal.sosfilt`.
+    phaseshift : (N,) numpy.ndarray
+        Phase shift in radians.
+
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        delay, weight, sos, phaseshift, selection, secondary_source = \
+            sfs.time.drivingfunction.nfchoa_25d_plane(array.x, R, npw, fs)
+        d = sfs.time.drivingfunction.nfchoa_25d_driving_signals(
+                delay, weight, sos, phaseshift, signal)
+        plot(d, selection, secondary_source)
+
+    """
+    if max_order is None:
+        max_order = util.max_order_circular_harmonics(len(x0))
+    if c is None:
+        c = default.c
+
+    x0 = util.asarray_of_rows(x0)
+    npw = util.asarray_1d(npw)
+    phi0, _, _ = util.cart2sph(*x0.T)
+    phipw, _, _ = util.cart2sph(*npw)
+    phaseshift = phi0 - phipw + np.pi
+
+    delay = -r0 / c
+    weight = 2
+    sos = []
+    for m in range(max_order + 1):
+        _, p, _ = besselap(m, norm='delay')
+        s_zeros = np.zeros(m)
+        s_poles = c / r0 * p
+        s_gain = 1
+        z_zeros, z_poles, z_gain = s2z(s_zeros, s_poles, s_gain, fs)
+        sos.append(zpk2sos(z_zeros, z_poles, z_gain, pairing='nearest'))
+    selection = util.source_selection_all(len(x0))
+    return delay, weight, sos, phaseshift, selection, secondary_source_point(c)
+
+
+def nfchoa_25d_point(x0, r0, xs, fs, max_order=None, c=None, s2z=matchedz_zpk):
+    r"""Virtual Point source by 2.5-dimensional NFC-HOA.
+
+    .. math::
+
+        D(\phi_0, s) =
+        \frac{1}{2\pi r_\text{s}}
+        \e{\frac{s}{c}(r_0-r_\text{s})}
+        \sum_{m=-M}^{M}
+        \Big(\frac{s-\frac{c}{r_\text{s}}\sigma_0}{s-\frac{c}{r_0}\sigma_0}\Big)^\mu
+        \prod_{l=1}^{\nu}
+        \frac{(s-\frac{c}{r_\text{s}}\sigma_l)^2-(\frac{c}{r_\text{s}}\omega_l)^2}
+        {(s-\frac{c}{r_0}\sigma_l)^2+(\frac{c}{r_0}\omega_l)^2}
+        \e{\i m(\phi_0 - \phi_\text{s})}
+
+    The driving function is represented in the Laplace domain,
+    from which the recursive filters are designed.
+    :math:`\sigma_l + \i\omega_l` denotes the complex roots of
+    the reverse Bessel polynomial.
+    The number of second-order sections is
+    :math:`\nu = \big\lfloor\tfrac{|m|}{2}\big\rfloor`,
+    whereas the number of first-order section :math:`\mu` is either 0 or 1
+    for even and odd :math:`|m|`, respectively.
+
+    Parameters
+    ----------
+    x0 : (N, 3) array_like
+        Sequence of secondary source positions.
+    r0 : float
+        Radius of the circular secondary source distribution.
+    xs : (3,) array_like
+        Virtual source position.
+    fs : int
+        Sampling frequency in Hertz.
+    max_order : int, optional
+        Ambisonics order.
+    c : float, optional
+        Speed of sound in m/s.
+    s2z : callable, optional
+        Function transforming s-domain poles and zeros into z-domain,
+        e.g. :func:`matchedz_zpk`, :func:`scipy.signal.bilinear_zpk`.
+
+    Returns
+    -------
+    delay : float
+        Overall delay in seconds.
+    weight : float
+        Overall weight.
+    sos : list of numpy.ndarray
+        Second-order section filters :func:`scipy.signal.sosfilt`.
+    phaseshift : (N,) numpy.ndarray
+        Phase shift in radians.
+
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        delay, weight, sos, phaseshift, selection, secondary_source = \
+            sfs.time.drivingfunction.nfchoa_25d_point(array.x, R, xs, fs)
+        d = sfs.time.drivingfunction.nfchoa_25d_driving_signals(
+                delay, weight, sos, phaseshift, signal)
+        plot(d, selection, secondary_source, t=ts)
+
+    """
+    if max_order is None:
+        max_order = util.max_order_circular_harmonics(len(x0))
+    if c is None:
+        c = default.c
+
+    x0 = util.asarray_of_rows(x0)
+    xs = util.asarray_1d(xs)
+    phi0, _, _ = util.cart2sph(*x0.T)
+    phis, _, rs = util.cart2sph(*xs)
+    phaseshift = phi0 - phis
+
+    delay = (rs - r0) / c
+    weight = 1 / 2 / np.pi / rs
+    sos = []
+    for m in range(max_order + 1):
+        _, p, _ = besselap(m, norm='delay')
+        s_zeros = c / rs * p
+        s_poles = c / r0 * p
+        s_gain = 1
+        z_zeros, z_poles, z_gain = s2z(s_zeros, s_poles, s_gain, fs)
+        sos.append(zpk2sos(z_zeros, z_poles, z_gain, pairing='nearest'))
+    selection = util.source_selection_all(len(x0))
+    return delay, weight, sos, phaseshift, selection, secondary_source_point(c)
+
+
+def nfchoa_3d_plane(x0, r0, npw, fs, max_order=None, c=None, s2z=matchedz_zpk):
+    r"""Virtual plane wave by 3-dimensional NFC-HOA.
+
+    .. math::
+
+        D(\phi_0, s) =
+        \frac{\e{\frac{s}{c}r_0}}{r_0}
+        \sum_{n=0}^{N}
+        (-1)^n (2n+1) P_{n}(\cos\Theta)
+        \Big(\frac{s}{s-\frac{c}{r_0}\sigma_0}\Big)^\mu
+        \prod_{l=1}^{\nu}
+        \frac{s^2}{(s-\frac{c}{r_0}\sigma_l)^2+(\frac{c}{r_0}\omega_l)^2}
+
+    The driving function is represented in the Laplace domain,
+    from which the recursive filters are designed.
+    :math:`\sigma_l + \i\omega_l` denotes the complex roots of
+    the reverse Bessel polynomial.
+    The number of second-order sections is
+    :math:`\nu = \big\lfloor\tfrac{|m|}{2}\big\rfloor`,
+    whereas the number of first-order section :math:`\mu` is either 0 or 1
+    for even and odd :math:`|m|`, respectively.
+    :math:`P_{n}(\cdot)` denotes the Legendre polynomial of degree :math:`n`,
+    and :math:`\Theta` the angle between :math:`(\theta, \phi)`
+    and :math:`(\theta_\text{pw}, \phi_\text{pw})`.
+
+    Parameters
+    ----------
+    x0 : (N, 3) array_like
+        Sequence of secondary source positions.
+    r0 : float
+        Radius of the spherical secondary source distribution.
+    npw : (3,) array_like
+        Unit vector (propagation direction) of plane wave.
+    fs : int
+        Sampling frequency in Hertz.
+    max_order : int, optional
+        Ambisonics order.
+    c : float, optional
+        Speed of sound in m/s.
+    s2z : callable, optional
+        Function transforming s-domain poles and zeros into z-domain,
+        e.g. :func:`matchedz_zpk`, :func:`scipy.signal.bilinear_zpk`.
+
+    Returns
+    -------
+    delay : float
+        Overall delay in seconds.
+    weight : float
+        Overall weight.
+    sos : list of numpy.ndarray
+        Second-order section filters :func:`scipy.signal.sosfilt`.
+    phaseshift : (N,) numpy.ndarray
+        Phase shift in radians.
+
+    """
+    if max_order is None:
+        max_order = util.max_order_spherical_harmonics(len(x0))
+    if c is None:
+        c = default.c
+
+    x0 = util.asarray_of_rows(x0)
+    npw = util.asarray_1d(npw)
+    phi0, theta0, _ = util.cart2sph(*x0.T)
+    phipw, thetapw, _ = util.cart2sph(*npw)
+    phaseshift = np.arccos(np.dot(x0 / r0, -npw))
+
+    delay = -r0 / c
+    weight = 4 * np.pi / r0
+    sos = []
+    for m in range(max_order + 1):
+        _, p, _ = besselap(m, norm='delay')
+        s_zeros = np.zeros(m)
+        s_poles = c / r0 * p
+        s_gain = 1
+        z_zeros, z_poles, z_gain = s2z(s_zeros, s_poles, s_gain, fs)
+        sos.append(zpk2sos(z_zeros, z_poles, z_gain, pairing='nearest'))
+    selection = util.source_selection_all(len(x0))
+    return delay, weight, sos, phaseshift, selection, secondary_source_point(c)
+
+
+def nfchoa_3d_point(x0, r0, xs, fs, max_order=None, c=None, s2z=matchedz_zpk):
+    r"""Virtual point source by 3-dimensional NFC-HOA.
+
+    .. math::
+
+        D(\phi_0, s) =
+        \frac{\e{\frac{s}{c}(r_0-r_\text{s})}}{4 \pi r_0 r_\text{s}}
+        \sum_{n=0}^{N}
+        (2n+1) P_{n}(\cos\Theta)
+        \Big(\frac{s-\frac{c}{r_\text{s}}\sigma_0}{s-\frac{c}{r_0}\sigma_0}\Big)^\mu
+        \prod_{l=1}^{\nu}
+        \frac{(s-\frac{c}{r_\text{s}}\sigma_l)^2-(\frac{c}{r_\text{s}}\omega_l)^2}
+        {(s-\frac{c}{r_0}\sigma_l)^2+(\frac{c}{r_0}\omega_l)^2}
+
+    The driving function is represented in the Laplace domain,
+    from which the recursive filters are designed.
+    :math:`\sigma_l + \i\omega_l` denotes the complex roots of
+    the reverse Bessel polynomial.
+    The number of second-order sections is
+    :math:`\nu = \big\lfloor\tfrac{|m|}{2}\big\rfloor`,
+    whereas the number of first-order section :math:`\mu` is either 0 or 1
+    for even and odd :math:`|m|`, respectively.
+    :math:`P_{n}(\cdot)` denotes the Legendre polynomial of degree :math:`n`,
+    and :math:`\Theta` the angle between :math:`(\theta, \phi)`
+    and :math:`(\theta_\text{s}, \phi_\text{s})`.
+
+    Parameters
+    ----------
+    x0 : (N, 3) array_like
+        Sequence of secondary source positions.
+    r0 : float
+        Radius of the spherial secondary source distribution.
+    xs : (3,) array_like
+        Virtual source position.
+    fs : int
+        Sampling frequency in Hertz.
+    max_order : int, optional
+        Ambisonics order.
+    c : float, optional
+        Speed of sound in m/s.
+    s2z : callable, optional
+        Function transforming s-domain poles and zeros into z-domain,
+        e.g. :func:`matchedz_zpk`, :func:`scipy.signal.bilinear_zpk`.
+
+    Returns
+    -------
+    delay : float
+        Overall delay in seconds.
+    weight : float
+        Overall weight.
+    sos : list of numpy.ndarray
+        Second-order section filters :func:`scipy.signal.sosfilt`.
+    phaseshift : (N,) numpy.ndarray
+        Phase shift in radians.
+
+    """
+    if max_order is None:
+        max_order = util.max_order_spherical_harmonics(len(x0))
+    if c is None:
+        c = default.c
+
+    x0 = util.asarray_of_rows(x0)
+    xs = util.asarray_1d(xs)
+    phi0, theta0, _ = util.cart2sph(*x0.T)
+    phis, thetas, rs = util.cart2sph(*xs)
+    phaseshift = np.arccos(np.dot(x0 / r0, xs / rs))
+
+    delay = (rs - r0) / c
+    weight = 1 / r0 / rs
+    sos = []
+    for m in range(max_order + 1):
+        _, p, _ = besselap(m, norm='delay')
+        s_zeros = c / rs * p
+        s_poles = c / r0 * p
+        s_gain = 1
+        z_zeros, z_poles, z_gain = s2z(s_zeros, s_poles, s_gain, fs)
+        sos.append(zpk2sos(z_zeros, z_poles, z_gain, pairing='nearest'))
+    selection = util.source_selection_all(len(x0))
+    return delay, weight, sos, phaseshift, selection, secondary_source_point(c)
+
+
+def nfchoa_25d_driving_signals(delay, weight, sos, phaseshift, signal):
+    """Get 2.5-dimensional NFC-HOA driving signals.
+
+    Parameters
+    ----------
+    delay : float
+        Overall delay in seconds.
+    weight : float
+        Overall weight.
+    sos : list of array_like
+        Second-order section filters :func:`scipy.signal.sosfilt`.
+    phaseshift : (N,) array_like
+        Phase shift in radians.
+    signal : (L,) array_like + float
+        Excitation signal consisting of (mono) audio data and a sampling
+        rate (in Hertz).  A `DelayedSignal` object can also be used.
+
+    Returns
+    -------
+    `DelayedSignal`
+        A tuple containing the delayed signals (in a `numpy.ndarray`
+        with shape ``(L, N)``), followed by the sampling rate (in Hertz)
+        and a (possibly negative) time offset (in seconds).
+
+    """
+    data, fs, t_offset = util.as_delayed_signal(signal)
+    N = len(phaseshift)
+    out = np.tile(np.expand_dims(sosfilt(sos[0], data), 1), (1, N))
+    for m in range(1, len(sos)):
+        modal_response = sosfilt(sos[m], data)[:, np.newaxis]
+        out += modal_response * np.cos(m * phaseshift)
+    return util.DelayedSignal(2 * weight * out, fs, t_offset + delay)
+
+
+def nfchoa_3d_driving_signals(delay, weight, sos, phaseshift, signal):
+    """Get 3-dimensional NFC-HOA driving signals.
+
+    Parameters
+    ----------
+    delay : float
+        Overall delay in seconds.
+    weight : float
+        Overall weight.
+    sos : list of array_like
+        Second-order section filters :func:`scipy.signal.sosfilt`.
+    phaseshift : (N,) array_like
+        Phase shift in radians.
+    signal : (L,) array_like + float
+        Excitation signal consisting of (mono) audio data and a sampling
+        rate (in Hertz).  A `DelayedSignal` object can also be used.
+
+    Returns
+    -------
+    `DelayedSignal`
+        A tuple containing the delayed signals (in a `numpy.ndarray`
+        with shape ``(L, N)``), followed by the sampling rate (in Hertz)
+        and a (possibly negative) time offset (in seconds).
+
+    """
+    data, fs, t_offset = util.as_delayed_signal(signal)
+    N = len(phaseshift)
+    out = np.tile(np.expand_dims(sosfilt(sos[0], data), 1), (1, N))
+    for m in range(1, len(sos)):
+        modal_response = sosfilt(sos[m], data)[:, np.newaxis]
+        out += (2 * m + 1) * modal_response * legendre(m, np.cos(phaseshift))
+    return util.DelayedSignal(weight / 4 / np.pi * out, fs, t_offset + delay)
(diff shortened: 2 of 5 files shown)


sfs/time/drivingfunction.py
```diff
@@ -251,7 +252,191 @@ def apply_delays(signal, delays):
...
+    for m in range(1, max_order+1):
+        modal_response = sig.sosfilt(sos[np.abs(m)], signal)
+        for l in range(N):
+            d[:,l] += modal_response * 2 * np.cos(m*phaseshift[l])
+    t_offset = delay[0]
+    return np.real(d) * weight, fs, t_offset
+    
+def _max_order_circular_harmonics(N, max_order):
+    """Compute order of 2D HOA."""
+    return (N-1) // 2 if max_order is None else max_order
```
hagenw: This is one of the functions we need for the time domain and the mono-frequent implementation. It is already implemented  [mono/drivingfunction.py](https://github.com/sfstoolbox/sfs-python/blob/master/sfs/mono/drivingfunction.py#L678), but in a slight different way:
```python
def _max_order_circular_harmonics(N, max_order):
    return N // 2 if max_order is None else max_order
```
Having a look at Ahrens (2012), p. 132 there should also be a distinction of cases for even and odd numbers:
```
     / N/2 - 1,   even nls
M = <
     \ (N-1)/2    odd nls
```

Another problem is that this only holds for 2D and 2.5D. For 3D we would have:
```
      _____
M = \|nls/2
```
see also [nfchoa_order.m](https://github.com/sfstoolbox/sfs-matlab/blob/master/SFS_general/nfchoa_order.m).
As we don't have 3D at the moment we could stick with the function name and change it later if we need 3D.

TODOS:
* find a common place and define only one `_max_order_circular_harmonics(N, max_order)` function
* add distinction between odd and even N
* add a reference to the help message

sfs/time/drivingfunction.py
```diff
@@ -251,7 +252,191 @@ def apply_delays(signal, delays):
...
+    delay : float
+        Overall delay (common to all secondary source)
+    weight : float
+        Overall weight (common to all secondary sources)
+    sos : dictionary
+        Second-order section filters 
+    phaseshift : float
+        Phase shift
+
+    """
```
hagenw: Add a reference to the help message. In sfs-matlab we used eq. (11) from Spors, Kuscher, Ahrens (2011).

sfs/time/drivingfunction.py
```diff
@@ -251,7 +252,191 @@ def apply_delays(signal, delays):
...
```
hagenw: Add a reference to the help message. In sfs-matlab we used eq. (10) from Spors, Kuscher, Ahrens (2011).

sfs/time/drivingfunction.py
hagenw: Hi @narahahn,
I just realized that `(N-1) // 2` is already the correct implementation of the max order function.
And the 2D vs. 3D problem can be solved by naming the 3D case `_max_order_spherical_harmonics`.
From the mentioned TODOs, then only these remain:
* find a common place and define only one `_max_order_circular_harmonics(N, max_order)` function
* add a reference to the help message

sfs/time/drivingfunction.py
```diff
@@ -251,7 +252,191 @@ def apply_delays(signal, delays):
...
     delays_samples -= offset_samples
-    out = np.zeros((delays_samples.max() + len(data), len(delays_samples)))
-    for column, row in enumerate(delays_samples):
-        out[row:row + len(data), column] = data
-    return util.DelayedSignal(out, samplerate, offset_samples / samplerate)
+    out = np.zeros((delays_samples.max() + len(signal), len(delays_samples)))
+    for channel, cdelay in enumerate(delays_samples):
+        out[cdelay:cdelay + len(signal), channel] = signal
+    return out, offset_samples / fs
+
```
chris-hld: This will break the current implementation of the time based WFS driving functions.
I think we wanted to try out using the `signal` class, here is an example :  https://github.com/sfstoolbox/sfs-python/blob/master/sfs/util.py#L141

narahahn: Thank you. Probably I chose the wrong script while merging.

sfs/time/drivingfunction.py
narahahn: Thank you for the review.
The references are now added with the respective equation numbers.

sfs/time/drivingfunction.py
narahahn: I agree. Maybe in `sfs.util`?

hagenw: Yes, I think `sfs.util` would be currently the best fit.

sfs/time/drivingfunction.py
```diff
@@ -6,6 +6,7 @@
 from __future__ import division
 import numpy as np
 from numpy.core.umath_tests import inner1d  # element-wise inner product
+import scipy.signal as sig
```
mgeier: I'm not sure if `sig` is a good name.
Typically, one would use `from scipy import signal`, but of course in our case we are already using `signal` as variable name ...

What about importing the individual functions by name?

    from scipy.signal import besselap, sosfilt, zpk2sos

mgeier:
```
$ python3 -m pycodestyle sfs/time/drivingfunction.py
sfs/time/drivingfunction.py:261:80: E501 line too long (84 > 79 characters)
sfs/time/drivingfunction.py:284:37: W291 trailing whitespace
sfs/time/drivingfunction.py:287:1: W293 blank line contains whitespace
sfs/time/drivingfunction.py:293:1: W293 blank line contains whitespace
sfs/time/drivingfunction.py:297:20: E225 missing whitespace around operator
sfs/time/drivingfunction.py:305:1: W293 blank line contains whitespace
sfs/time/drivingfunction.py:327:80: E501 line too long (83 > 79 characters)
sfs/time/drivingfunction.py:350:37: W291 trailing whitespace
sfs/time/drivingfunction.py:354:1: W293 blank line contains whitespace
sfs/time/drivingfunction.py:360:1: W293 blank line contains whitespace
sfs/time/drivingfunction.py:364:20: E225 missing whitespace around operator
sfs/time/drivingfunction.py:388:32: E231 missing whitespace after ','
sfs/time/drivingfunction.py:388:37: E231 missing whitespace after ','
sfs/time/drivingfunction.py:388:39: E231 missing whitespace after ','
sfs/time/drivingfunction.py:394:80: E501 line too long (93 > 79 characters)
sfs/time/drivingfunction.py:404:37: W291 trailing whitespace
sfs/time/drivingfunction.py:421:20: E225 missing whitespace around operator
sfs/time/drivingfunction.py:429:20: E231 missing whitespace after ','
sfs/time/drivingfunction.py:429:23: E231 missing whitespace after ','
sfs/time/drivingfunction.py:433:12: E231 missing whitespace after ','
sfs/time/drivingfunction.py:437:16: E231 missing whitespace after ','
sfs/time/drivingfunction.py:440:1: W293 blank line contains whitespace
sfs/time/drivingfunction.py:441:1: E302 expected 2 blank lines, found 1
sfs/time/drivingfunction.py:445:1: E302 expected 2 blank lines, found 1
sfs/time/drivingfunction.py:457:1: W391 blank line at end of file
```

Those are not all in your changes, I'm just trying to say that you should use some kind of a linter, ideally as an editor plugin or something.

sfs/time/drivingfunction.py
mgeier: While you are at it, you could replace the `if`/`else` expression with an `if`/`else` statement, I think this would be easier to understand.

Then it should also become more obvious that you should remove the separate check for `None` before calling the function.

sfs/time/drivingfunction.py
narahahn: I agree. Now only those three functions are imported. Thanks!

sfs/time/drivingfunction.py
hagenw: I tried to implement this with ef78d76, where I moved `_max_order_circular_harmonics` to `util`,  renamed it to `max_order_circular_harmonics` and added documentation.

For the documentation I would need some help. At the moment I added `page 132 of [Ahrens2012]_`, but it would be nicer to replace this by the actual equation. As I don't have a copy of the book here I cannot do this myself.

mgeier:
> The SOS filter coefficients are stored in a dictionary [...]. Is there a more elegant way?

Will there ever be some orders missing?

If not, I think a simple `list` would be better, because it can simply be indexed from 0 to the maximum order.
And it can be easily iterated in a well-defined order.

But if there is ever an order missing (or not starting from 0), this wouldn't work. Then a `dict` would make sense IMHO.

narahahn:
@mgeier Thank you, it works very well. Just simple.
And I don't think we'll ever use missing-order-Ambisonics :-)

examples/time_domain_nfchoa.py
```diff
@@ -0,0 +1,71 @@
...
+plt.axis('tight')
+plt.colorbar()
+plt.clim([-120, 0])
+
+plt.figure()
+plt.imshow(np.real(d), interpolation='None', cmap='coolwarm')
+plt.axis('tight')
+plt.colorbar()
+plt.clim([-0.1, 0.1])
+
```
mgeier: The above two figures are not saved, is this intentional?

examples/time_domain_nfchoa.py
```diff
@@ -0,0 +1,71 @@
...
+
+# Synthesized sound field
+a0 = np.ones(len(x0))
```
mgeier: Why do you overwrite `a0` here?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
     for column, row in enumerate(delays_samples):
         out[row:row + len(data), column] = data
     return util.DelayedSignal(out, samplerate, offset_samples / samplerate)
+
+
+def nfchoa_25d_plane(x0, r0, npw, max_order=None, c=None, fs=44100, normalize=True):
```
mgeier: Is it possible to avoid the explicit specification of a default sampling rate?

We've avoided that up to now, and it would be nice to keep it that way.

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+    x0 : (N, 3) array_like
+        Sequence of secondary source positions.
+    r0 : float
+        Radius of the circular secondary source distribution
+    npw : (3,) array_like
+        Unit vector (propagation direction) of plane wave.
+    max_order : int
+        Ambisonics order
+    c : float, optional
+        Speed of sound
```
mgeier: `fs` and `normalize` are not documented

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+
+    Returns
+    -------
+    delay : float
+        Overall delay (common to all secondary source)
```
mgeier: in seconds?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+    weight : float
+        Overall weight (common to all secondary sources)
+    sos : list
+        Second-order section filters
+    phaseshift : float
+        Phase shift
```
mgeier: in radians?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+    c : float, optional
+        Speed of sound
+
+    Returns
+    -------
+    delay : float
+        Overall delay (common to all secondary source)
+    weight : float
+        Overall weight (common to all secondary sources)
+    sos : list
```
mgeier: You could probably write `list of array_like`.

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+        Second-order section filters
```
mgeier: What about a link to the SciPy docs where the data format is explained?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+    phaseshift : float
+        Phase shift
+
+    References
+    ----------
+        S. Spors, V. Kuscher, J. Ahrens (2011) - "Efficient realization of
+       model-based rendering for 2.5-dimensional near-field compensated higher
+       order Ambisonics", WASPAA, p. 61-64
+
+       See Eq.(10)
```
mgeier: Please move the reference to the reference page.

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+
+    x0 = util.asarray_of_rows(x0)
+    npw = util.asarray_1d(npw)
+    phipw, _, _ = util.cart2sph(*npw)
+    phi0, _, _ = util.cart2sph(*x0.T)
+
+    delay = -r0/c
+    weight = 2
+    sos = []
+    for m in range(0, max_order+1):
```
mgeier: It's not necessary to specify `0`.

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
     for column, row in enumerate(delays_samples):
         out[row:row + len(data), column] = data
     return util.DelayedSignal(out, samplerate, offset_samples / samplerate)
+
+
+def nfchoa_25d_plane(x0, r0, npw, max_order=None, c=None, fs=44100, normalize=True):
+    """Vritual plane wave by 2.5-dimensional NFC-HOA.
```
mgeier: typo

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+
+    delay = -r0/c
+    weight = 2
+    sos = []
+    for m in range(0, max_order+1):
+        _, p, _ = besselap(m, norm='delay')
+        # TODO: modify "besselap" for very high orders (>150)
+        s0 = np.zeros(m)
+        sinf = (c/r0)*p
+        z0 = np.exp(s0/fs)
```
mgeier: It's unclear to me why you create a bunch of zeros and then divide them by something.
And the exponential of that will always be 1, won't it?
This can't be right?!?

mgeier: BTW, would it be appropriate to copy the relevant equations from the paper to the docstring?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+            k = _normalize_digital_filter_gain(s0, sinf, z0, zinf, fs=fs)
+        else:
+            k = 1
+        sos.append(zpk2sos(z0, zinf, k, pairing='nearest'))
+        # TODO: normalize the SOS filters individually?
+    phaseshift = phipw + np.pi - phi0
+    return delay, weight, sos, phaseshift
+
+
+def nfchoa_25d_point(x0, r0, xs, max_order=None, c=None, fs=44100, normalize=True):
```
mgeier: Again, it would be nice to avoid an explicit default sampling rate.

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+
+    Parameters
+    ----------
+    x0 : (N, 3) array_like
+        Sequence of secondary source positions.
+    r0 : float
+        Radius of the circular secondary source distribution
+    npw : (3,) array_like
+        Unit vector (propagation direction) of plane wave.
+    max_order : int
```
mgeier: `int, optional`

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+
+    Parameters
+    ----------
+    x0 : (N, 3) array_like
+        Sequence of secondary source positions.
+    r0 : float
+        Radius of the circular secondary source distribution
+    xs : (3,) array_like
+        Virtual source position.
+    max_order : int
```
mgeier: `int, optional`

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+        Ambisonics order
+    c : float, optional
+        Speed of sound
```
mgeier: `fs` and `normalize` is undocumented.

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+
+    Returns
+    -------
+    delay : float
+        Overall delay (common to all secondary source)
```
mgeier: seconds?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+    weight : float
+        Overall weight (common to all secondary sources)
+    sos : list
```
mgeier: same as above or reference to the above function.

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+        Second-order section filters
+    phaseshift : float
+        Phase shift
```
mgeier: radians?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+
+    References
+    ----------
+        S. Spors, V. Kuscher, J. Ahrens (2011) - "Efficient realization of
+       model-based rendering for 2.5-dimensional near-field compensated higher
+       order Ambisonics", WASPAA, p. 61-64
+
+       See Eq.(11)
```
mgeier: references page

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+            k = _normalize_digital_filter_gain(s0, sinf, z0, zinf, fs=fs)
+        else:
+            k = 1
+        sos.append(zpk2sos(z0, zinf, k, pairing='nearest'))
+        # TODO: normalize the SOS filters individually?
+    phaseshift = phi0 - phi
+    return delay, weight, sos, phaseshift
+
+
+def nfchoa_driving_signals(delay, weight, sos, phaseshift, signal, max_order=None, fs=44100):
```
mgeier: Can you please use the `signal` argument like in `driving_signals()` above?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+        Second-order section filters
+    phaseshift : (C,) array_like
+        Phase shift
+    signal : (N,) array_like
+        Excitation signal (mono) which gets weighted and delayed.
+    fs: int, optional
+        Sampling frequency in Hertz.
+
+    Returns
+    -------
```
mgeier: I think it would make sense to return a `DelayedSignal`, like in `driving_signals()`.

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+    driving_signals : (N, C) numpy.ndarray
+        Driving signals.
+    t_offset : float
+        Simulation point in time offset (seconds).
+
+    """
+    max_order = util.max_order_circular_harmonics(len(phaseshift), max_order)
+
+    delay = util.asarray_1d(delay)
+    weight = util.asarray_1d(weight)
```
mgeier: Why are you doing this? Shouldn't those be scalars?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+    d = np.zeros((L, N), dtype='complex128')
+
+    modal_response = sosfilt(sos[0], signal)
+    for l in range(N):
+        d[:, l] += modal_response
+    for m in range(1, max_order+1):
+        modal_response = sosfilt(sos[np.abs(m)], signal)
+        for l in range(N):
+            d[:, l] += modal_response * 2 * np.cos(m*phaseshift[l])
+    t_offset = delay[0]
```
mgeier: Again, shouldn't this be a scalar?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,194 @@ def apply_delays(signal, delays):
...
+    return np.real(d) * weight, fs, t_offset
+
+
+def _normalize_digital_filter_gain(s0, sinf, z0, zinf, fs=44100):
+    """Match the digital filter gain at the Nyquist freuqneycy"""
```
mgeier: typo

sfs/util.py
```diff
@@ -466,3 +466,34 @@ def _count_walls_1d(a):
         order = order[max_order_mask, :]
         
     return xs, order
+
+
+def max_order_circular_harmonics(N, max_order):
```
mgeier: I think `max_order` should have a default of `None`.

But I'm not really sure if that function signature makes sense anyway ...

The "hidden" check for `None` seems a bit weird to me.

What about checking for `None` each time before use?

```python
if max_order is None:
    max_order = util.max_order_circular_harmonics(len(x0))
```

What do you think?

sfs/util.py
```diff
@@ -466,3 +466,34 @@ def _count_walls_1d(a):
...
+    Parameters
+    ----------
+    N : int
+        Number of secondary sources.
+    max_order : int
+        Ambisonics order
+
+    Returns
+    -------
+    max_order : int
```
mgeier: There's no need to specify a variable name if there is only one output.

In fact, the whole `Returns` section could be removed.

mgeier:
```
$ python3 -m pydocstyle sfs/time/drivingfunction.py 
sfs/time/drivingfunction.py:261 in public function `nfchoa_25d_plane`:
        D413: Missing blank line after last section ('References')
sfs/time/drivingfunction.py:326 in public function `nfchoa_25d_point`:
        D207: Docstring is under-indented
sfs/time/drivingfunction.py:326 in public function `nfchoa_25d_point`:
        D413: Missing blank line after last section ('References')
sfs/time/drivingfunction.py:438 in private function `_normalize_digital_filter_gain`:
        D202: No blank lines allowed after function docstring (found 1)
sfs/time/drivingfunction.py:438 in private function `_normalize_digital_filter_gain`:
        D400: First line should end with a period (not 'y')
```

sfs/time/drivingfunction.py
hagenw: The extra reference pages was not part of this branch before. I merged it with the latest version of master and you can find the reference page at `doc/references.rst`.

examples/time_domain_nfchoa.py
narahahn: I think driving functions plots are not that necessary. I'll remove them. It was just for visual inspection.

By the way, do we have to save all the generated figures?

examples/time_domain_nfchoa.py
narahahn: I thought that the weights are included in the driving functions, so I set `a0` to unity.
But I was wrong, we do need `2 * np.pi * R / N` for spatial discretization.

sfs/time/drivingfunction.py
narahahn: Sure

sfs/time/drivingfunction.py
narahahn: OK.

I will remove the `normalize` option since it does not make sense not normalizing the filter. The option was used to demonstrate the effect of normalization in the [AES paper](http://www.aes.org/e-lib/browse.cfm?elib=18609).

sfs/time/drivingfunction.py
narahahn: Yes, thank you.

sfs/time/drivingfunction.py
narahahn: Do you mean the url like [this](https://docs.scipy.org/doc/scipy-0.19.1/reference/generated/scipy.signal.sosfilt.html#scipy.signal.sosfilt)?

sfs/time/drivingfunction.py
narahahn: You're right. `z0` will be just ones. Will be fixed.

I just wanted to make it easier to understand what's going on (at least for myself). The poles/zeros in the Laplace domain are mapped to the z-domain using the matches z-transform. I thought this might also help reviewing the code.

narahahn: As far as I know, the equations for pole/zero mapping are not in our papers. Maybe we can find it in DSP or signal processing textbooks.

sfs/time/drivingfunction.py
narahahn: Do you need the equation number? It's (4.26)

sfs/util.py
```diff
@@ -479,14 +479,45 @@ def _count_walls_1d(a):
     wall_count = np.concatenate([_count_walls_1d(d) for d in xs.T], axis=1)
     xs *= L
 
-    if prune is True:
+     if prune is True:
```
mgeier: IndentationError: unexpected indent

examples/time_domain_nfchoa.py
mgeier: Well, the idea of the example scripts is that they generate PNG files for the plots, isn't it?
If you don't use `savefig()`, there is no visible result ...

BTW, @hagenw has just merged the `nbsphinx` stuff into `master`. If you like, you can change your example from a script into a Jupyter notebook?

sfs/time/drivingfunction.py
mgeier: yes

sfs/time/drivingfunction.py
mgeier: OK, if you want to show this, you should add a comment. Otherwise it looks like a mistake.

Or you could make a separate (private?) function with a meaningful name that should also make clear what's happening?

You could probably re-use such a function in `nfchoa_25d_point()`?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,205 @@ def apply_delays(signal, delays):
...
+    d = np.zeros((L, N))
+#    TODO : check FOS/SOS structure
+
+    d = np.tile(np.expand_dims(sosfilt(sos[0], data), 1), (1, N))
+    for m in range(1, max_order+1):
+        modal_response = sosfilt(sos[np.abs(m)], data)
+        for n in range(N):
+            d[:, n] += modal_response * 2 * np.cos(m*phaseshift[n])
+    t_offset += delay
+    return np.real(d) * weight, fs, t_offset
```
mgeier: This should return a `DelayedSignal`, right?

And why is `np.real()` used here?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,205 @@ def apply_delays(signal, delays):
...
+    -------
+    driving_signals : (N, C) numpy.ndarray
+        Driving signals.
+    t_offset : float
+        Simulation point in time offset (seconds).
+
+    """
+    max_order = util.max_order_circular_harmonics(len(phaseshift), max_order)
+    data, fs, t_offset = util.as_delayed_signal(signal)
+    N = len(phaseshift)
```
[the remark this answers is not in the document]
narahahn: My mistake, thanks.

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,205 @@ def apply_delays(signal, delays):
...
+    L = len(data)
```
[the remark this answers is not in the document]
narahahn: Will be fixed, thanks.

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,205 @@ def apply_delays(signal, delays):
...
+    d = np.zeros((L, N))
+#    TODO : check FOS/SOS structure
+
+    d = np.tile(np.expand_dims(sosfilt(sos[0], data), 1), (1, N))
+    for m in range(1, max_order+1):
+        modal_response = sosfilt(sos[np.abs(m)], data)
```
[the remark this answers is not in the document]
narahahn: You're right.
It's a trace of my old implementation where I iterated `m` from `-max_order` to `max_order`

sfs/time/drivingfunction.py
narahahn: This is also something I need before when the `modal_response` was multiplied with a complex term.

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,205 @@ def apply_delays(signal, delays):
...
+    """
+    max_order = util.max_order_circular_harmonics(len(phaseshift), max_order)
+    data, fs, t_offset = util.as_delayed_signal(signal)
+    N = len(phaseshift)
+    L = len(data)
+    d = np.zeros((L, N))
+#    TODO : check FOS/SOS structure
+
+    d = np.tile(np.expand_dims(sosfilt(sos[0], data), 1), (1, N))
+    for m in range(1, max_order+1):
```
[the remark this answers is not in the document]
narahahn: Yes, it's the same as `len(sos)-1`

sfs/util.py
```diff
@@ -487,6 +487,32 @@ def _count_walls_1d(a):
...
+    If ``max_order`` is ``None`` it returns the maximum order
+    for which no spatial aliasing appears.
+    It is given on page 132 of [Ahrens2012]_ as
+
+    .. math::
+        \text{max_order} =
+            \begin{cases}
+                N/2 - 1 & \text{even}\;N \\
+                (N-1)/2 & \text{odd}\;N
+            \end{cases}
```
mgeier: Since you are using a different (but apparently equivalent) equation in the code, you should mention this here.

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,205 @@ def apply_delays(signal, delays):
...
+
+    Returns
+    -------
+    delay : float
+        Overall delay in seconds.
+    weight : float
+        Overall weight.
+    sos : list of `numpy.ndarray`
+        Second-order section filters :func:`scipy.signal.sosfilt`.
+    phaseshift : (N,) array_like
```
mgeier: This should be `` `numpy.ndarray` `` instead of `array_like`?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,205 @@ def apply_delays(signal, delays):
...
```
mgeier: Again,  `` `numpy.ndarray` `` ?

sfs/time/drivingfunction.py
```diff
@@ -255,3 +256,205 @@ def apply_delays(signal, delays):
...
+def nfchoa_driving_signals(delay, weight, sos, phaseshift, signal):
+    """Get NFC-HOA driving signals.
+
+    Parameters
+    ----------
+    delay : float
+        Overall delay in seconds.
+    weight : float
+        Overall weight.
+    sos : list of `numpy.ndarray`
```
mgeier: This should be "list of array_like".

sfs/util.py
```diff
@@ -487,6 +487,32 @@ def _count_walls_1d(a):
...
+    which is equivalent to
+
+    .. math::
+        \text{max_order} = \big\lfloor \frac{N - 1}{2} \big\rfloor.
+
+    Parameters
+    ----------
+    N : int
+        Number of secondary sources.
+    """
```
mgeier: Missing blank line.

mgeier:
Thanks for the updates!

I've marked all my outdated requests as "resolved" and added a few more comments.

Apart from this, is this PR ready for merging?

If yes, can you please squash all commits and rebase to `master`?

fs446:
For nfchoa_25d_point() and all other related time domain NFC-HOA (and maybe somewhere else?)
the current usage denotation of s0, sinf, z0, zinf
`k = _normalize_digital_filter_gain(s0, sinf, z0, zinf, fs)`
might be changed to 
`k = _normalize_digital_filter_gain(szero, spole, zzero, zpole, fs)`
for improved readability. Although coding becomes slightly longer the consistency, self-explanation and same-length convention could be helpful.

narahahn:
`_matchedz_zpk()` now computes the gain `zgain` of the SOS filters.
This makes the behavior consistant with `bilinear_zpk()`,
in terms of input and output variables.

Variable names `sgain` and `zgain` are used instead of `k`
for the system gain in the s- and z-domain, respectively.
So the system function is described by {`szero`, `spole`, `sgain`},
and {`zzero`, `zpole`, `zgain`}.

The gain normalization is now performed inside `_matchedz_zpk()`.
This makes `_normalize_digital_filter_gain()` obsolete.
It will be removed after some tests.

mgeier:
I agree that `zzero` is better than `z0`, but what about `z_zero`?
I think it would be more readable with an underscore.

Also, shouldn't it be plural?

`z_zeros`, `z_poles`, `s_zeros`, `s_poles`, `z_gain`, `s_gain`

> This makes the behavior consistant with `bilinear_zpk()`,
in terms of input and output variables.

That's great! Now you can get rid of the ugly strings and just pass functions around.
You should also change `_matchedz_zpk()` to `matchedz_zpk()` for that.

This, in turn, makes the appendix `_method` obsolete (and slightly wrong).

The signature could then look like this:

```python
def nfchoa_25d_point(x0, r0, xs, fs, max_order=None, c=None, s2z=matchedz_zpk):
```

narahahn:
> I agree that `zzero` is better than `z0`, but what about `z_zero`?
> I think it would be more readable with an underscore.

I also prefer using underscore,
but then I would like to propose yet another convention:

`zdomain_zeros`, `sdomain_poles`

which in my opinion reads better and is more self explanatory.
The single letter `z` might be misunderstood as zero (like in `zpk`)
or the z-coordinate.

> That's great! Now you can get rid of the ugly strings and just pass functions around.

Great idea, thanks!
Then we don't need the if-else statement
and can just call either function by `s2k(z, p, k, fs)`, right?

narahahn:
> The signature could then look like this:
> 
> ```python
> def nfchoa_25d_point(x0, r0, xs, fs, max_order=None, c=None, s2z=matchedz_zpk):
> ```

Hmm ... This doesn't work because `matchedz_zpk` and `bilinear_zpk`
are not defined in the current workspace where  `nfchoa_25d_point()` is called.
Any idea?

mgeier:
> `zdomain_zeros`, `sdomain_poles`

Well at some point we'll have to stop making the names longer ...

This would have to be `z_domain_zeros` to be consistent, and the two underscores make it a bit tedious in my opinion. But I think it would still be acceptable.

If you prefer it, that's fine for me.

> The single letter `z` might be misunderstood as zero (like in `zpk`)
or the z-coordinate.

That's a risk that I would be willing to take.

> Then we don't need the if-else statement and can just call either function by `s2k(z, p, k, fs)`, right?

Exactly.

> Hmm ... This doesn't work because `matchedz_zpk` and `bilinear_zpk`
are not defined in the current workspace where  `nfchoa_25d_point()` is called.

Do you mean where it is *called* or where it is *defined*?

The default argument must be available where the function is *defined*. That means that you'll have to move the definition of `matchedz_zpk()` further up in the file.
Or, probably better, move it to an entirely different place (e.g. `sfs.util`?) and import it in `drivingfunction.py`.

OTOH, `bilinear_zpk` doesn't have to be defined anywhere. It's the responsibility of a user to import it from `scipy.signal`.
It would probably make sense to mention this in the documentation.

narahahn:
> This would have to be `z_domain_zeros` to be consistent, and the two underscores make it a bit tedious in my opinion. But I think it would still be acceptable.

We are already using variable/function names like
`samplerate`, `drivingfunction`, `soundfield`, `wfs_25d_preeq`.
I think `zdomain_` isn't so inconsistent.
BTW, I wouldn't use two underscores, either.

> The default argument must be available where the function is _defined_. That means that you'll have to move the definition of `matchedz_zpk()` further up in the file.
> Or, probably better, move it to an entirely different place (e.g. `sfs.util`?) and import it in `drivingfunction.py`.

Although passing the functions is very a fancy idea,
it could be a bit inconvenient or even confusing
for non-expert users like me.

From
```python
def nfchoa_25d_point(x0, r0, xs, fs, max_order=None, c=None, s2z=matchedz_zpk):
```
I would expect that 
```python
nfchoa_25d_point(x0, r0, xs, fs, max_order, c):
```
and
```python
nfchoa_25d_point(x0, r0, xs, fs, max_order, c, s2z=matchedz_zpk)
```
are the same, but actually the second one will work only if
`matchedz_zpk` is defined where the driving function is called.
```python
from sfs.util import matchedz_zpk
d = sfs.time.drivingfunction.nfchoa_25d_point(
    x0, r0, xs, fs, max_order, c, s2z=matchedz_zpk)
```
or
```python
d = sfs.time.drivingfunction.nfchoa_25d_point(
    x0, r0, xs, fs, max_order, c, s2z=sfs.time.drivingfunction.matched_zpk)
```

mgeier:
> I think `zdomain_` isn't so inconsistent.

It is inconsistent in the sense that the first two words are directly joined together, while the second and third word are joined by an underscore.

I think we should choose one or the other.

But PEP-8 doesn't say anything about this: https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

If you really prefer `zdomain_zeros`, I'm fine with it.
I find it awkward, but I can live with it.

> but actually the second one will work only if
`matchedz_zpk` is defined where the driving function is called.

Is it also confusing that `x0`, `r0`, `xs` and `fs` have to be defined where the driving function is called?

This is just how Python works: names have to be defined.

And function definition may happen in a different namespace than function use.

Users will have to know a few Python feature in order to use the SFS toolbox. And if they don't know those features, we can give them an opportunity to learn, which is great!

narahahn:
3D NFC-HOA driving functions for virtual plane waves and point sources are added.
Synthesized sound fields of virtual plane waves are shown below.

pw_soundfield

pw_soundfield

The source signals are band-limited impulses (22 kHz and 1 kHz).
240 secondary sources are placed on a sphere with R=1.5 m.
I used pre-calculated t-design distribution from
http://neilsloane.com/sphdesigns/dim3/

narahahn:
Equations describing the NFC-HOA driving functions are added in the docstring (c600808).
Although the equations and the respective functions are
intended for time-domain implementations,
the driving functions are represented in the Laplace domain
from which the filter coefficients are computed.
However, this looks somewhat inconsistent considering that
the WFS driving functions are all represented in the time-domain.
We may evaluate the inverse Laplace transform and use the
time-domain representation, but IMO, this would be misleading
since we never use it.
Do you have any thoughts?

@mgeier 
Can I use an in-line LaTeX sytax in the docstring?
I want to define some variables appearing in the equations.

mgeier:
> Can I use an in-line LaTeX sytax in the docstring?

Yes you can and you should!

You can use inline math with ``:math:`x` `` and display math with

```rst
.. math::

    x
```

Just don't forget to use a raw string `r"""..."""` if you use backslashes (which I guess you'll do).

narahahn:
> You can use inline math with `` :math:`x` `` and display math with
> 
> ```
> .. math::
> 
>     x
> ```

Great! Thank you.

sfs/time/drivingfunction.py
```diff
@@ -296,7 +296,18 @@ def matchedz_zpk(s_zero, s_pole, s_gain, fs):
...
+    .. math::
+
+        D(\phi_0, s) =
+        2e^{\frac{s}{c}r_0}
+        \sum_{m=-M}^{M}
+        (-1)^m
+        \Big(\frac{s}{s-\frac{c}{r_0}\sigma_0}\Big)^\mu
+        \prod_{l=1}^{\nu}
+        \frac{s^2}{(s-\frac{c}{r_0}\sigma_l)^2+(\frac{c}{r_0}\omega_l)^2}
+        e^{im(\phi_0 - \phi_\text{pw})}
```
hagenw: You could use `\i` instead of `i` and `\e{...}` instead of `e^{...}` which will result in non italic symbols as proposed original by @fs446.
For an example result see the documentation of `sfs.mono.drivingfunction.nfchoa_25d_plane`

narahahn: Thank you for the review. The math symbols are changed now.

sfs/time/drivingfunction.py
```diff
@@ -254,3 +256,422 @@ def apply_delays(signal, delays):
...
+        Radius of the circular secondary source distribution.
+    xs : (3,) array_like
+        Virtual source position.
+    fs : int
+        Sampling frequency in Hertz.
+    max_order : int, optional
+        Ambisonics order.
+    c : float, optional
+        Speed of sound in m/s.
+    s2z : function
```
mgeier: Same as above, I'd prefer `callable`.

And it's `optional`.

sfs/time/drivingfunction.py
```diff
@@ -254,3 +256,422 @@ def apply_delays(signal, delays):
...
+        Radius of the circular secondary source distribution.
+    npw : (3,) array_like
+        Unit vector (propagation direction) of plane wave.
+    fs : int
+        Sampling frequency in Hertz.
+    max_order : int, optional
+        Ambisonics order.
+    c : float, optional
+        Speed of sound in m/s.
+    s2z : function
```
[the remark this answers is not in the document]
mgeier: Also, the argument is `optional`.

sfs/time/drivingfunction.py
```diff
@@ -254,3 +256,422 @@ def apply_delays(signal, delays):
...
+        Function transforming s-domain poles and zeros into z-domain,
+        e.g. :func:`matchedz_zpk`, :func:`scipy.signal.bilinear_zpk`.
+
+    Returns
+    -------
+    delay : float
+        Overall delay in seconds.
+    weight : float
+        Overall weight.
+    sos : list of array_like
```
mgeier: `list of numpy.ndarray`?

sfs/time/drivingfunction.py
```diff
@@ -254,3 +256,422 @@ def apply_delays(signal, delays):
...
+        Radius of the spherical secondary source distribution.
+    npw : (3,) array_like
+        Unit vector (propagation direction) of plane wave.
+    fs : int
+        Sampling frequency in Hertz.
+    max_order : int, optional
+        Ambisonics order.
+    c : float, optional
+        Speed of sound in m/s.
+    s2z : function
```
mgeier: Same as above: `callable` and `optional`.

sfs/time/drivingfunction.py
```diff
@@ -254,3 +256,422 @@ def apply_delays(signal, delays):
...
+        Function transforming s-domain poles and zeros into z-domain,
+        e.g. :func:`matchedz_zpk`, :func:`scipy.signal.bilinear_zpk`.
+
+    Returns
+    -------
+    delay : float
+        Overall delay in seconds.
+    weight : float
+        Overall weight.
+    sos : list of array_like
```
mgeier: `numpy.ndarray`

sfs/time/drivingfunction.py
```diff
@@ -254,3 +256,422 @@ def apply_delays(signal, delays):
...
+        Second-order section filters :func:`scipy.signal.sosfilt`.
+    phaseshift : (N,) array_like
```
mgeier: `numpy.ndarray`

sfs/time/drivingfunction.py
```diff
@@ -254,3 +256,422 @@ def apply_delays(signal, delays):
...
+        Radius of the spherial secondary source distribution.
+    xs : (3,) array_like
+        Virtual source position.
+    fs : int
+        Sampling frequency in Hertz.
+    max_order : int, optional
+        Ambisonics order.
+    c : float, optional
+        Speed of sound in m/s.
+    s2z : function
```
mgeier: `callable, optional`

sfs/time/drivingfunction.py
```diff
@@ -254,3 +256,422 @@ def apply_delays(signal, delays):
...
+        Function transforming s-domain poles and zeros into z-domain,
+        e.g. :func:`matchedz_zpk`, :func:`scipy.signal.bilinear_zpk`.
+
+    Returns
+    -------
+    delay : float
+        Overall delay in seconds.
+    weight : float
+        Overall weight.
+    sos : list of array_like
```
mgeier: `numpy.ndarray`

sfs/time/drivingfunction.py
```diff
@@ -254,3 +256,422 @@ def apply_delays(signal, delays):
...
+        Second-order section filters :func:`scipy.signal.sosfilt`.
+    phaseshift : (N,) array_like
```
mgeier: `numpy.ndarray`

narahahn: Thank you for the review!

doc/examples/time_domain_nfchoa.py
```diff
@@ -0,0 +1,58 @@
...
+signal = dirac, fs, 0
+
+# NFC-HOA order
+max_order = None
+
+# POINT SOURCE
+#xs = np.r_[1.5, 1.5, 0]  # position
+#delay, weight, sos, phaseshift = sfs.time.drivingfunction.nfchoa_25d_point(
+#                                     x0, R, xs, fs, max_order=max_order)
+#t = np.linalg.norm(xs) / sfs.defs.c
```
mgeier: We should try to avoid committing code that's commented out, see #108.

This is very hard to maintain, as exemplified here because the code is outdated (it uses `sfs.defs`).

Can you please either completely remove the code or comment it in and make an additional plot from it?

doc/examples/time_domain_nfchoa.py
```diff
@@ -0,0 +1,58 @@
...
+d, fs, t_offset = sfs.time.drivingfunction.nfchoa_25d_driving_signals(
+                      delay, weight, sos, phaseshift, signal)
+
+# Synthesized sound field
+p = sfs.time.soundfield.p_array(x0, (d, fs, t_offset), a0, t, grid)
+
+plt.figure()
+sfs.plot.level(p, grid, cmap='Blues')
+sfs.plot.loudspeaker_2d(x0, n0)
+#sfs.plot.virtualsource_2d(xs, type='point')
```
mgeier: Again, please no commented-out code.

doc/examples/time_domain_nfchoa.py
```diff
@@ -0,0 +1,58 @@
...
+sfs.plot.virtualsource_2d([0, 0], ns=npw, type='plane')
+plt.savefig('pw_level.png')
+
+plt.figure()
+sfs.plot.soundfield(p, grid, cmap='coolwarm')
+sfs.plot.loudspeaker_2d(x0, n0)
+#sfs.plot.virtualsource_2d(xs, type='point')
```
mgeier: No comment.

narahahn:
> Is it intentional that sequences of zeros and poles are called "zero" and "pole" in singular?

Sorry, I missed that. They're changed now.

narahahn:
> Thanks, that's great, I really like it.

I really appreciate your review and comments!

Merging from the `master` branch was really the worst idea.
Rewriting the history was quite painful.

> Now the only problem is that there were some API changes in `master` (mainly driving functions returning some more stuff) ... could you please update your PR accordingly?

The code are updated according to the recent changes.

But when I build the HTML pages, the math symbols like `\i`, `\e` are not properly displayed.
Same result in the `master` branch. Is there some work in progress that might cause this?

mgeier:
Thanks a lot @narahahn, I think it's time to merge this!

The documentation for the new return values is missing, but we can add this at a later time.

> Merging from the `master` branch was really the worst idea.

Well some people like this workflow and they keep all those merge commits in their history. I don't like this, because I think it makes the history basically useless.
I prefer doing `git rebase master` regularly, and when there are too many commits in the feature branch, squashing some of them with `git rebase --interactive`.

But anyway, now it's done, which is great!

> Is there some work in progress that might cause this?

I think this is a problem with your local files, I don't see this problem. You should probably remove the old "build" directory and try again.

PR #47, opened 2017-11-23
Generalization of the modal order argument

spors:
The argument `N` can now handle scalars or a vector containing for all three dimensions `None` for computation of max order, a scalar for a given maximum order or a vector for given orders per dimension. For instance

- `N=10 or N = [10, 10, 10]` computes the field in all three dimensions up to order 10
- `N = [None, None, [0]]` calculates the order in the first two dimensions and uses zeroth-order for the third dimension

### DIFF
--- a/sfs/mono/source.py
+++ b/sfs/mono/source.py
@@ -169,10 +169,9 @@ def point_modal(omega, x0, n0, grid, L, N=None, deltan=0, c=None):
     L : (3,) array_like
         Dimensionons of the rectangular room.
     N : (3,) array_like or int, optional
-        Modal orders in the three-spatial dimensions to calculate the sound
-        field for or maximum orders. If not given, the maximum modal order is
-        approximately determined and the sound field is computed up to this
-        maximum order. Scalars apply to all three dimensions.
+        For all three spatial dimensions per dimension maximum order,
+        list of orders used or if not given maximum modal order is
+        approximately determined. A scalar applies to all three dimensions.
     deltan : float, optional
         Absorption coefficient of the walls.
     c : float, optional
@@ -189,7 +188,7 @@ def point_modal(omega, x0, n0, grid, L, N=None, deltan=0, c=None):
     x, y, z = util.as_xyz_components(grid)
 
     if np.isscalar(N):
-        N = N * np.ones(3)
+        N = N * np.ones(3, dtype=int)
 
     if N is None:
             N = [None, None, None]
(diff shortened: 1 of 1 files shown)


sfs/mono/source.py
```diff
@@ -189,31 +188,30 @@ def point_modal(omega, x0, n0, grid, L, N=None, deltan=0, c=None):
     x0 = util.asarray_1d(x0)
     x, y, z = util.as_xyz_components(grid)
 
+    if np.isscalar(N):
+        N = N * np.ones(3)
```
mgeier: `np.ones()` creates floating point values by default.
You should use `dtype=int` to get integers.

Then you can remove the `int()` from line 204.

sfs/mono/source.py
```diff
@@ -169,11 +169,9 @@ def point_modal(omega, x0, n0, grid, L, N=None, deltan=0, c=None):
...
         Dimensionons of the rectangular room.
     N : (3,) array_like or int, optional
-        Combination of modal orders in the three-spatial dimensions to
-        calculate the sound field for or maximum order for all
-        dimensions.  If not given, the maximum modal order is
-        approximately determined and the sound field is computed up to
-        this maximum order.
+        For all three spatial dimensions per dimension maximum order,
+        list of orders used or if not given maximum modal order is
+        approximately determined. A scalar applies to all three dimensions.
```
hagenw: This is a little bit difficult and I would prefer to have the default behavior at the end, so maybe:
```
For all three spatial dimensions per dimension maximum order or
￼list of orders. A scalar applies to all three dimensions. If no
order is provided it is approximately determined. 
```

sfs/mono/source.py
```diff
@@ -189,31 +187,30 @@ def point_modal(omega, x0, n0, grid, L, N=None, deltan=0, c=None):
...
-        mm = [N[0]]
-        nn = [N[1]]
-        ll = [N[2]]
+            N = [None, None, None]
+
+    orders = [0, 0, 0]
+    for i in range(3):
+        if N[i] is None:
+            # compute max order
+            orders[i] = range(np.ceil(L[i]/np.pi * k) + 1)
```
hagenw: Could you provide a reference for this or is this to trivial?

mgeier: You should keep the conversion to `int` here, because `np.ceil()` returns a `float`.

spors: I just calculated it exploiting the fact that the wavenumber and angular frequency are connected to each other. However, this only applies to the case of rigid boundaries. As soon as `deltan` is not zero, a higher order would have to be used for a given accuracy due to the decay of the resonances.

spors: Should we add this information to the documentation as note?

hagenw: I would say yes. If you have interested students that look into it, it would be valuable information.

PR #48, opened 2017-11-27
DOC: Use nbsphinx

mgeier:
To get started, I've turned the "modal room acoustics" example into a Jupyter notebook.

I hope that many more notebooks will follow.


hagenw:
Cool.

We should check on readthedocs, but I guess we have to include #44 first, otherwise it [fails to build](https://readthedocs.org/projects/sfs-python/builds/6331961/).

Locally I was able to run it with `python setup.py build_sphinx`, but I had to install jupyter and pandoc first.

mgeier:
Yes, after #44 it should work.

And you are right, `pandoc` was missing, I added it in 9c48a2bcd06e5b4fbdfb64bfe247da524b5b85c4.

I hope it works like that ...

mgeier:
I've rebased this PR and now it works on RTD!

@hagenw Can you please merge if there are no objections?

hagenw:
Maybe you should update sphinx to `sphinx>=1.3.6` in `doc/readthedocs-environment.yml`. Otherwise the github link is broken as described in https://github.com/spatialaudio/nbsphinx/issues/26

mgeier:
I've changed to version 1.3.6 in 2c49ee21545269132e6f2ccbcbc9e1db5b3d8547, but it still doesn't work.
Can we move this to a separate issue?

hagenw:
Yes, I will create an issue after merging.
Would you like to incorporate 2c49ee2 or remove it before merging?

mgeier:
We can keep it, doesn't hurt anyone ...

hagenw:
Good news, the link is not broken:
http://python.sfstoolbox.org/en/latest/examples/modal-room-acoustics.html
It points to https://github.com/sfstoolbox/sfs-python//blob/master/doc/examples/modal-room-acoustics.ipynb, which is correct.

mgeier:
@hagenw Thanks for merging!

Great that the link works for "latest"!

However, there still seem to be problems because it doesn't work in the "nbsphinx" branch and I don't know if it will work if we make a new release.

For reference:
https://github.com/rtfd/readthedocs.org/issues/1820
https://github.com/rtfd/readthedocs.org/issues/1637

PR #42, opened 2017-09-29
DOC: Description of image source method

trettberg:
Mainly in `util.image_sources_for_box()`:

* description of source selection options

* IMHO misleading parameter names changed

PR #49, opened 2017-11-28
Update to references

hagenw:
This includes:
1.) Update the references to the source image model from #42 to be in line with #46 by moving them to the extra `doc/references.rst` file
2.) Cleaning up `doc/references.rst` by introducing line breaks
3.) Fixing a typo

### DIFF
--- a/doc/references.rst
+++ b/doc/references.rst
@@ -1,20 +1,56 @@
 References
 ==========
 
-.. [Ahrens2012] Ahrens, J., *Analytic Methods of Sound Field Synthesis*. (Springer, Berlin Heidelberg, 2012), doi:`10.1007/978-3-642-25743-8 <https://doi.org/10.1007/978-3-642-25743-8>`__
+.. [Ahrens2012]
+    Ahrens, J.,
+    *Analytic Methods of Sound Field Synthesis*.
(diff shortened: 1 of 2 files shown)


hagenw:
If you do a citation in a paper you would use "Name1" for one author "Name1 and Name2" for two authors and "Name1 et al." for three or more authors. That was the idea behind the referencing and it is the same as in SFS theory.

But I agree that most probably it would be better to use just one name here as I'm also not willing to add "etal" into the brackets. I will merge this first and then think again about it.

mgeier:
Yes, "Name1 and Name2 (2007)" or "(Name1 and Name2, 2007)" would be fine, but "[Name1Name22007]" is an entirely different thing.

Did you ever try https://github.com/mcmtroffaes/sphinxcontrib-bibtex?

hagenw:
I thought about it, but I never tried. It is also a little bit [buggy using multiple pages](http://sphinxcontrib-bibtex.readthedocs.io/en/latest/usage.html#unresolved-citations-across-documents), but I will give it a try.

mgeier:
I wasn't aware of the multi-page issue, but using `:all:` doesn't sound too bad, does it?
Why would we want to have unused entries in our bibtex file anyway?

PR #51, opened 2017-12-03
Use sphinxcontrib-bibtex for references

hagenw:
As proposed in #49 I switched to [sphinxcontrib-bibtex](https://github.com/mcmtroffaes/sphinxcontrib-bibtex/) for references.

Using the `unsrt` style this creates a numbered ([1], [2], ...) list of references, see http://python.sfstoolbox.org/en/bibtex/references.html.

This means inside the text we have now also only numbered links like "Eq.(17) from [[7]](http://python.sfstoolbox.org/en/bibtex/references.html#spors2008)", see [wfs_2d_plane](http://python.sfstoolbox.org/en/bibtex/frequency-domain.html#sfs.mono.drivingfunction.wfs_2d_plane). In most of the times this should be fine, I only changed it for the first entry in the [docstring of image_sources_for_box](http://python.sfstoolbox.org/en/bibtex/utilities.html#sfs.util.image_sources_for_box).

In a third commit I also started to clean up the links provided in the reference by replacing links to PDF versions of AES papers with the official permalinks. This is only optional and don't have to be included, but I thought it looks cleaner that way. The disadvantage is of course, that if you are not an AES member you have to google for the PDF.

### DIFF
--- a/doc/readthedocs-environment.yml
+++ b/doc/readthedocs-environment.yml
@@ -4,6 +4,7 @@ dependencies:
   - python==3.5
   - sphinx>=1.3.6
   - sphinx_rtd_theme
+  - sphinxcontrib-bibtex
   - numpy
   - scipy
   - matplotlib>=1.5
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -1,0 +1,72 @@
+@book{Ahrens2012,
+    author = {Ahrens, J.},
+    title = {{Analytic Methods of Sound Field Synthesis}},
+    publisher = {Springer},
+    address = {Berlin Heidelberg},
+    year = {2012},
+    doi = {10.1007/978-3-642-25743-8}
+}
+@book{Moser2012,
+    author = {M\"{o}ser, M.},
+    title = {{Technische Akustik}},
+    publisher = {Springer},
+    address = {Berlin Heidelberg},
+    year = {2012},
+    doi = {10.1007/978-3-642-30933-5}
+}
+@inproceedings{Spors2010,
+    author = {Spors, S. and Ahrens, J.},
+    title = {{Analysis and Improvement of Pre-equalization in 2.5-dimensional
+        Wave Field Synthesis}},
+    booktitle = {128th Convention of the Audio Engineering Society},
+    year = {2010},
+    url = {http://bit.ly/2Ad6RRR}
+}
+@inproceedings{Spors2009,
+    author = {Spors, S. and Ahrens, J.},
+    title = {{Spatial Sampling Artifacts of Wave Field Synthesis for the
+        Reproduction of Virtual Point Sources}},
+    booktitle = {126th Convention of the Audio Engineering Society},
+    year = {2009},
+    url = {http://bit.ly/2jkfboi}
+}
+@inproceedings{Spors2016,
+    author = {Spors, S. and Schultz, F. and Rettberg, T.},
+    title = {{Improved Driving Functions for Rectangular Loudspeaker Arrays
+        Driven by Sound Field Synthesis}},
+    booktitle = {42nd German Annual Conference on Acoustics (DAGA)},
+    year = {2016},
+    url = {http://bit.ly/2AWRo7G}
+}
+@inproceedings{Spors2008,
+    author = {Spors, S. and Rabenstein, R. and Ahrens, J.},
+    title = {{The Theory of Wave Field Synthesis Revisited}},
+    booktitle = {124th Convention of the Audio Engineering Society},
+    year = {2008},
+    url = {http://bit.ly/2ByRjnB}
+}
+@phdthesis{Wierstorf2014,
+    author = {Wierstorf, H.},
+    title = {{Perceptual Assessment of Sound Field Synthesis}},
+    school = {Technische Universität Berlin},
+    year = {2014},
+    doi = {10.14279/depositonce-4310}
+}
+@article{Allen1979,
+    author = {Allen, J. B. and  Berkley, D. A.},
+    title = {{Image method for efficiently simulating small‐room acoustics}},
+    journal = {Journal of the Acoustical Society of America},
+    volume = {65},
+    pages = {943--950},
+    year = {1979},
+    doi = {10.1121/1.382599}
+}
+@article{Borish1984,
+    author = {Borish, J.},
+    title = {{Extension of the image model to arbitrary polyhedra}},
+    journal = {Journal of the Acoustical Society of America},
+    volume = {75},
+    pages = {1827--1836},
+    year = {1984},
+    doi = {10.1121/1.390983}
+}
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -7,7 +7,7 @@
     doi = {10.1007/978-3-642-25743-8}
 }
 @book{Moser2012,
-    author = {M\"{o}ser, M.},
+    author = {Möser, M.},
     title = {{Technische Akustik}},
     publisher = {Springer},
     address = {Berlin Heidelberg},
--- a/doc/references.rst
+++ b/doc/references.rst
@@ -1,56 +1,6 @@
 References
 ==========
 
-.. [Ahrens2012]
-    Ahrens, J.,
-    *Analytic Methods of Sound Field Synthesis*.
-    (Springer, Berlin Heidelberg, 2012),
-    doi:`10.1007/978-3-642-25743-8 <https://doi.org/10.1007/978-3-642-25743-8>`__
-
-.. [Moser2012]
-    Möser, M.,
-    *Technische Akustik*.
-    (Springer, Berlin Heidelberg, 2012),
-    doi:`10.1007/978-3-642-30933-5 <https://doi.org/10.1007/978-3-642-30933-5>`__
-
-.. [SporsAhrens2010]
-    Spors, S., Ahrens, J.,
-    “Analysis and Improvement of Pre-equalization in 2.5-dimensional Wave Field Synthesis,”
-    in *128th Conv. Audio Eng. Soc.*, London, UK (2010), Paper 8121,
-    `pdf <https://www.int.uni-rostock.de/fileadmin/user_upload/publications/spors/2010/AES128_Spors_etal_WFS_preeq.pdf>`__
-
-.. [SporsAhrens2009]
-    Spors, S., Ahrens, J.,
-    “Spatial Sampling Artifacts of Wave Field Synthesis for the Reproduction of Virtual Point Sources,”
-    in *126th Conv. Audio Eng. Soc.*, Munich, Germany (2009), Paper 7744,
-    `pdf <https://www.int.uni-rostock.de/fileadmin/user_upload/publications/spors/2009/AES126_Spors_etal_WFS_spatial_aliasing.pdf>`__
-
-.. [Spors2016]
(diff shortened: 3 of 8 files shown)


doc/readthedocs-environment.yml
```diff
@@ -4,6 +4,7 @@ dependencies:
   - python==3.5
   - sphinx>=1.3.6
   - sphinx_rtd_theme
+  - sphinxcontrib-bibtex
```
mgeier: You should also add this to `doc/requirements.txt`.

doc/references.bib
```diff
@@ -0,0 +1,80 @@
...
+    year = {2010},
+    url = {http://www.aes.org/e-lib/browse.cfm?elib=15418}
+}
+@inproceedings{Spors2009,
+    author = {Spors, S. and Ahrens, J.},
+    title = {{Spatial Sampling Artifacts of Wave Field Synthesis for the
+        Reproduction of Virtual Point Sources}},
+    booktitle = {126th Conv. Audio Eng. Soc.},
+    address = {Munich, Germany},
+    ages = {Paper 7744},
```
mgeier: typo

doc/references.rst
```diff
@@ -1,56 +1,6 @@
...
-    J. Acoust. Soc. Am. **65**, 943-950 (1979),
-    doi:`10.1121/1.382599 <https://doi.org/10.1121/1.382599>`__
-
-.. [Borish1984]
-    Borish, J.,
-    “Extension of the image model to arbitrary polyhedra,”
-    J. Acoust. Soc. Am. **75**, 1827-1836 (1984),
-    doi:`10.1121/1.390983 <https://doi.org/10.1121/1.390983>`__
+.. bibliography:: references.bib
+    :style: plain
```
mgeier: Are numbered references really the best choice here?
What about using the default `alpha` style?

mgeier:
Cool, that works nicely!

mgeier:
Is it really necessary to abbreviate journal and conference names?

Is it really necessary to provide the locations of conferences?
I find this a bit childish.

> replacing links to PDF versions of AES papers with the official permalinks

If would be even nicer to have DOIs, but it looks like they don't exist for AES Convention papers?

Having the direct PDF links was definitely useful, can't we have both somehow?

doc/references.bib
```diff
@@ -0,0 +1,80 @@
+@book{Ahrens2012,
+    author = {Ahrens, J.},
+    title = {{Analytic Methods of Sound Field Synthesis}},
+    publisher = {Springer},
+    address = {Berlin Heidelberg},
+    year = {2012},
+    doi = {10.1007/978-3-642-25743-8}
+}
+@book{Moser2012,
+    author = {M\"{o}ser, M.},
```
mgeier: If you switch to the `alpha` style, you should write the proper letter `ö` here (and probably use the UTF-8 setting for the `bibliography` directive). Otherwise it doesn't seem to be parsed correctly.

doc/readthedocs-environment.yml
hagenw: I added it, but I'm wondering why we have both files?
I thought `requirements.txt` is needed if we use pip and `readthedocs-environment.yml` if we use conda, which we are doing at the moment, aren't we?

doc/references.rst
hagenw: At the moment there are only two choices with sphinxcontrib-bibtex, one is the numbered version and the other one is `alpha`. The latter one I dislike very much as it is neither the proper names of the authors nor an ordered list of references, but it's also fine with me to switch to that one.
if we stick with `plain` we have to remember that we have to use [a], [b], ... for footnotes instead of [1], [2], ....

hagenw:
> Is it really necessary to abbreviate journal and conference names?

I prefer this as it provides shorter reference entries.

> Is it really necessary to provide the locations of conferences?
> I find this a bit childish.
> Is the paper number really necessary?

For some journals they force you to include those and I find it nice if you can just reuse our bibtex entries for this, but it's not strictly necessary. But we should agree on one approach to be consistent with all entries.

>> replacing links to PDF versions of AES papers with the official permalinks

>If would be even nicer to have DOIs, but it looks like they don't exist for AES Convention papers?

That's correct, some conferences like AES and DAGA don't provide DOIs. Otherwise we could always provide the DOIs.

> Having the direct PDF links was definitely useful, can't we have both somehow?

I guess the only way is to modify the plugin for this. What you could do in Bibtex is to use the `note` entry to add anything you want, for example:
```bibtex
@inproceedings{Spors2008,
    author = {Spors, S. and Rabenstein, R. and Ahrens, J.},
    title = {{The Theory of Wave Field Synthesis Revisited}},
    booktitle = {124th Conv. Audio Eng. Soc.},
    address = {Amsterdam, Netherlands},
    pages = {Paper 7358},
    year = {2008},
    note = {{URL: http://www.aes.org/e-lib/browse.cfm?elib=14488} and 
        \url{http://bit.ly/SRA08pdf}}
}
```
but this translates into non click-able links in the output:
url1

If we use the PDF links, I would be in favour of shortening them, e.g. using bit.ly as you get ugly problems on smaller screens with those long links:
url2

doc/readthedocs-environment.yml
mgeier: You are right, we are using `conda` on RTD and we don't need `requirements.txt` for that.
But it is still useful for users like me who want to install the necessary packages locally with

    python3 -m pip install -r doc/requirements.txt

like it's mentioned in http://python.sfstoolbox.org/en/latest/contributing.html

doc/references.rst
mgeier: Of course it would be nicest to have full author names and years, but I think that's unrealistic right now.

I guess we'll have to choose the lesser of two evils.
If you prefer the numbered style, I'm fine with it.

mgeier:
>> Is it really necessary to abbreviate journal and conference names?
>
> I prefer this as it provides shorter reference entries.

But why?
I understand it when there is an extra cost for each page of a journal publication, but in our case there is no limit.

And I'm quite sure that the abbreviations actually need more time and cognitive effort to read than the full names.

>> Is it really necessary to provide the locations of conferences?
>> I find this a bit childish.
>> Is the paper number really necessary?
>
> For some journals they force you to include those

Which journals?
I'm having a hard time believing that any serious journal would impose such a ridiculous rule.

> and I find it nice if you can just reuse our bibtex entries for this, but it's not strictly necessary.

Which entries are you talking about reusing?

> But we should agree on one approach to be consistent with all entries.

Yes, definitely, that's why I brought it up.

I've tried the `note` entry, but sadly that's not processed as reST, which I think would be a good feature for `sphinxcontrib-bibtex`.

Since it seems that for now we can provide only one URL, I think I'd prefer the more useful PDF link.

And I think shortening the URLs is a good idea.

doc/readthedocs-environment.yml
hagenw: That's true, and I'm using exactly this for building it locally.
Is there an advantage of using conda on RTD?

doc/references.rst
hagenw: I'm still not completely decided as I also see some advantages of `alpha`.
I will do a few more tests and then propose what I would prefer.

hagenw:
>>> Is it really necessary to abbreviate journal and conference names?
>>
>> I prefer this as it provides shorter reference entries.
>
>But why?
>I understand it when there is an extra cost for each page of a journal publication, but in our case there is no limit.
>
>And I'm quite sure that the abbreviations actually need more time and cognitive effort to read than the full names.

Got that point, I think it might be also less work for us as we don't have to find the abbreviation.

>>> Is it really necessary to provide the locations of conferences?
>>> I find this a bit childish.
>>> Is the paper number really necessary?
>>
>> For some journals they force you to include those
>
>Which journals?
>I'm having a hard time believing that any serious journal would impose such a ridiculous rule.

For example, Jasa, Springer and IEEE were strict with those (at least with the paper number when there are no pages). 

>> and I find it nice if you can just reuse our bibtex entries for this, but it's not strictly necessary.
>
>Which entries are you talking about reusing?

The entries from the bibtex file for writing a paper.

hagenw:
>> But we should agree on one approach to be consistent with all entries.
>
>Yes, definitely, that's why I brought it up.

OK, here a summary on what we agreed so far and which points are still open:

- [x] Include full Journal/Conference name
- [x] Include DOI
- [x] If available include link to PDF of paper as `url`. Shorten that link using e.g. bit.ly.
- [x] If no DOI and no PDF, but a permalink to the paper (e.g. http://www.aes.org/e-lib/browse.cfm?elib=14940) is available use this as `url`
- [ ] include addresses for conferences
- [ ] include paper numbers for AES convention/conference papers
- [ ] if no DOI is given for a book, include ISBN

doc/references.rst
hagenw: I have switched now to `alpha`. It is indeed not perfect, but you can still use the abbreviation to some amount like a name. I tested this also in the theory repository.

If we are motivated we could at a later stage have a look if we would like to add another option to sphinxcontrib-bibtex itself.

doc/readthedocs-environment.yml
mgeier: The answer to this question keeps changing over time ...

In the past, there were problems with a broken `pandoc` installation on the default RTD deployment (without `conda`). And the versions of installed packages like e.g. `matplotlib` were hopelessly out-of-date (see #2).

Some issues that were solved by switching to `conda` but should (probably?) have been solved in the meantime:
https://github.com/spatialaudio/nbsphinx/commit/e10d0d8c7708607297f67e94bde505cd142d29f1#diff-c10bbd14be638b56c95217c4943387ff
https://github.com/spatialaudio/nbsphinx/issues/69
https://github.com/rtfd/readthedocs.org/issues/2521

The overall situation may have improved in the meantime.

I would actually prefer to switch away from `conda`, but in the past it just made less problems ...

doc/references.rst
mgeier: OK, cool. And yes, it would be great if this would be added to `sphinxcontrib-bibtex`, but I'm not going to work on that ...

mgeier:
Thanks for the update!

I still find the paper number superfluous and the conference locations ridiculously superfluous and dubious. But if you like them, please keep them, it's not such a big deal.

Using the ISBN if no DOI is available sounds reasonable.

I also agree with dotted first names.
In an actual LaTeX context, I'd write full first names and full middle names (if known) into the `.bib` file and use the BibTeX/BibLaTeX style to do the abbreviations.
But this doesn't seem to be possible with `sphinxcontrib-bibtex`.
This is another reason why `.bib` files often cannot be re-used without changes.

LGTM!

hagenw:
As you are probably right and it will not be easy with the reuse of the bibtex file, I finally decided to remove address and paper number as there is not much usage for them then.
I'm going to merge this now and summarize on what we agreed on here, if anybody is looking this up later on:

- Include full Journal/Conference name
- Include DOI
- If available include link to PDF of paper as `url`. Shorten that link using e.g. bit.ly.
- If no DOI and no PDF, but a permalink to the paper (e.g. http://www.aes.org/e-lib/browse.cfm?elib=14940) is available use this as `url`
- DON'T include addresses for conferences
- DON'T include paper numbers for AES convention/conference papers
- If no DOI is given for a book, include ISBN
- In `author` use abbreviation of first name(s) with dot afterwards: `{Name1, I. N. and Name2, A.}`

PR #52, opened 2017-12-04
DOC: add config files for binder

mgeier:
The interactive example notebook provided by this PR is currently available at:

https://mybinder.org/v2/gh/sfstoolbox/sfs-python/binder?filepath=doc/examples

The name of the current branch is "binder"; when this PR is merged this should be replaced by "master".

As a bonus feature, I've also added a new reST role that creates a link to a notebook using the current Git commit. This way, the documentation can link to a binder instance that runs the exact same version of the SFS toolbox.


hagenw:
Cool to have the interactive version as well.
I tested it and everything works fine.

doc/examples/index.rst
```diff
@@ -11,6 +11,10 @@ Various examples are located in the directory ``doc/examples/``, e.g.
     Illustrates the synthesis of sound figures with Wave Field Synthesis
 
 
+The Jupyter notebooks are also available online as interactive examples:
+:binder:`doc/examples`.
```
hagenw: Which Jupyter notebooks?
The overall example page is a little bit messy at the moment. At the top we introduce it with "Various examples are located in the directory doc/examples/, e.g.", but we don't mention that some are just python code you have to execute yourself and others are python notebooks.

Personally, I would prefer to have a link to the interactive notebook inside the sphinx notebook site itself (e.g. http://python.sfstoolbox.org/en/latest/examples/modal-room-acoustics.html) rather then one generic link on the first page. Is this possible?

mgeier: > Which Jupyter notebooks?

Those that are in `doc/examples/`.
Currently it's only one.

> The overall example page is a little bit messy at the moment.

Yes, it is totally just a raw draft right now.

The whole documentation is.

I just wanted to include an example for the ``:binder:`...` `` syntax somewhere.

We should definitely add more text at some point.
But I also hope that most of the Python scripts will be turned into notebooks.
Only the computationally intensive ones should stay in Python scripts, in order  not to exceed the 15 minutes time limit on readthedocs.org.

Feel free to add more text and more notebooks!

This PR is just about enabling more content, not about the content itself ...

> I would prefer to have a link to the interactive notebook inside the sphinx notebook site itself

Yes this would be great.

But I think it's not possible to have a link to the correct binder version that also works seamlessly in Jupyter itself.
OTOH, if the notebook is running in Jupyter already, you wouldn't need a link to binder anyway.

I think it would be ideal if we could add some generic text and a few links automatically to the top/bottom/sidebar of all notebooks. Namely a link to the Github location, and a link to binder. Both should link to the right commit, not just to "master".
AFAIK, that's currently not possible with `nbsphinx`, there is a related issue: https://github.com/spatialaudio/nbsphinx/issues/5.

hagenw: Ok, that makes sense.
Maybe you could change the text to
```rst
Examples
========

Various examples are located in the directory ``doc/examples/`` as
python scripts, e.g.

* sound_field_synthesis.py:
    Illustrates the general usage of the toolbox
* horizontal_plane_arrays.py:
    Computes the sound fields for various techniques, virtual sources and loudspeaker array configurations
* soundfigures.py:
    Illustrates the synthesis of sound figures with Wave Field Synthesis


Or Jupyter notebooks, which are also available online as interactive examples:
:binder:`doc/examples`.


.. toctree::
    :maxdepth: 1

    modal-room-acoustics
```

mgeier: done: e3b9651498c172b8c88ff67eba784c65659fca68

hagenw: Cool, feel free to merge it.

mgeier: @hagenw 

> I would prefer to have a link to the interactive notebook inside the sphinx notebook site itself

I've created https://github.com/spatialaudio/nbsphinx/pull/150 which can be used for exactly that, without "polluting" the notebooks themselves.

What do you think about it?

PR #53, opened 2017-12-15
Update to new sfstoolbox.org/en/3.0 permalinks

hagenw:
This upgrades all the permalinks to equations on http://sfstoolbox.org to the new http://sfstoolbox.org/en/3.0 version.


hagenw:
This is obsolete, given #103.

PR #54, opened 2018-01-10
DOC: add link to theory, matlab, and python to sidebar

hagenw:
To make it more obvious that we three different sub-projects under SFS Toolbox I created a small navigation on the right side. You can preview it at http://python.sfstoolbox.org/en/nav/

sfs-python

A few remarks:
1) The last commit has to be deleted before merging
2) The related pull requests https://github.com/sfstoolbox/theory/pull/25 and https://github.com/sfstoolbox/sfs-matlab/pull/180 have to be merged as well
3) For the matlab and python version we also have to make an official release as the default url points to their last stable release

### DIFF
--- a/doc/_static/css/title.css
+++ b/doc/_static/css/title.css
@@ -1,0 +1,33 @@
+.wy-side-nav-search>a, .wy-side-nav-search .wy-dropdown>a {
+    font-family: "Roboto Slab","ff-tisa-web-pro","Georgia",Arial,sans-serif;
+    font-size: 200%;
+    margin-top: .222em;
+    margin-bottom: .202em;
+}
+.wy-side-nav-search {
+    padding: 0;
+}
+form#rtd-search-form {
+    margin-left: .809em;
+    margin-right: .809em;
+}
+.rtd-nav a {
+    float: left;
+    display: block;
+    width: 33.3%;
+    height: 100%;
+    padding-top: 7px;
+    color: white;
+}
+.rtd-nav {
+    overflow: hidden;
+    width: 100%;
+    height: 35px;
+    margin-top: 15px;
+}
+.rtd-nav a:hover {
+    background-color: #388bbd;
+}
+.rtd-nav a.active {
+    background-color: #388bbd;
+}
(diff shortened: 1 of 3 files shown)


hagenw:
Interesting. I guess the problem is an outdated build as the current build of http://python.sfstoolbox.org/en/nav/ is 1 month old and it works for sfs-matlab:
http://matlab.sfstoolbox.org/en/nav/search/?q=normalization

I tried to rebuild the nav branch of sfs-python on RTD, but it fails with an nbsphinx error message:
```
Exception occurred:
  File "/home/docs/checkouts/readthedocs.org/user_builds/sfs-python/conda/nav/lib/python3.5/site-packages/nbsphinx.py", line 1291, in depart_code_latex
    assert False
AssertionError
```
See: https://readthedocs.org/projects/sfs-python/builds/6537037/

I tried it with version 0.2.18 of nbsphinx as well (which works on my local PC), but get the same error on RTD: https://readthedocs.org/projects/sfs-python/builds/6537313/

Any ideas how to fix this?

mgeier:
OK, this is due to a change in Sphinx 1.6.6 and I forgot to change `nbsphinx` accordingly. I've created a PR now: https://github.com/spatialaudio/nbsphinx/pull/159.

With that, the build succeeds: https://readthedocs.org/projects/sfs-python/builds/6538817/

I'll merge this soon and I'll make a new `nbsphinx` release in the next few days.

hagenw:
Cool, I wait until then and update they build afterwards.

mgeier:
I've just released `nbsphinx` and rebuilt the docs of the `nav` branch, now the search functionality works!

hagenw:
Cool, thank you.

I have updated how the actual path for the Python tab is set. Before it was hard coded now it is determined by `pathto(master_doc)`, which is also used to get the url for the clickable title (SFS Toolbox) of the page. This has the advantage that it will work independent of the current branch.
For the Matlab and Theory tab this can of course not be done and we have to add a fixed url. But of course inside those projects it can be handled the same way.

mgeier:
Sounds good!

hagenw:
I would like to merge this one, but think that it would be nice to have a release shortly afterwards as the default url (http://python.sfstoolbox.org) always points to the latest release, which is the old web page without a new release.

mgeier:
We can make a release any time.

PR #57, opened 2018-02-26
Clip soundfield into float64 range for plotting

### LINKED ISSUE #50: Monofrequent dipole source example plot
The plot of the mono-frequent dipole source has a (normalization?) issue in the current version.

frequency-domain-9_new

http://python.sfstoolbox.org/en/latest/frequency-domain.html#sfs.mono.source.point_dipole

Before it was

frequency-domain-9_old

http://python.sfstoolbox.org/en/0.3.1/#sfs.mono.source.point_dipole

trettberg:
closes #50 

(This may become obsolete in matplotlib 2.2 or not.
Since it's easy enough, I propose to fix it now anyway.)

### DIFF
--- a/sfs/plot.py
+++ b/sfs/plot.py
@@ -285,6 +285,7 @@ def soundfield(p, grid, xnorm=None, cmap='coolwarm_clip', vmin=-2.0, vmax=2.0,
     if ax is None:
         ax = plt.gca()
 
+    # see https://github.com/matplotlib/matplotlib/issues/10567
     if matplotlib_version.startswith('2.1.'):
         p = np.clip(p, -1e15, 1e15)  # clip to float64 range
 
--- a/sfs/plot.py
+++ b/sfs/plot.py
@@ -1,6 +1,7 @@
 """Plot sound fields etc."""
 from __future__ import division
 import matplotlib.pyplot as plt
+from matplotlib import __version__ as matplotlib_version
 from matplotlib.patches import PathPatch
 from matplotlib.path import Path
 from matplotlib.collections import PatchCollection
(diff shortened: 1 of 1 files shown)


hagenw:
I'm also in favor of including a fix for this.
An alternative solution would be to restrict matplotlib to <=2.0.2 in the dependencies of sfs and change this to >=2.2 when they have fixed matplotlib. I'm not sure what would be the better option.

mgeier:
I wouldn't restrict the `matplotlib` version, as this could be an unnecessary inconvenience for users.
BTW, currently we don't even officially depend on `matplotlib` (according to `setup.py`).

The fix suggested by @trettberg seems less intrusive, but what about limiting the fix to the affected versions (since it looks like it'll be fixed in 2.2.0)?
Something like:

```python
import matplotlib
if matplotlib.__version__.startswith('2.1'):
    p = np.clip(p, -1e15, 1e15)
```

sfs/plot.py
```diff
@@ -284,6 +285,9 @@ def soundfield(p, grid, xnorm=None, cmap='coolwarm_clip', vmin=-2.0, vmax=2.0,
     if ax is None:
         ax = plt.gca()
 
+    if matplotlib_version.startswith('2.1.'):
+        p = np.clip(p, -1e15, 1e15)  # clip to float64 range
```
hagenw: Maybe we add a link to the matplotlib bug (https://github.com/matplotlib/matplotlib/issues/10567) on the line before to highlight why we are doing this?

hagenw: Otherwise this implementation is fine with me.

PR #60, opened 2018-06-18
pytest for cart2sph & sph2cart

trojanjay:


chohner:
Two thoughts:
-  You will probably want to use `numpy.testing.assert_allclose()` (https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.testing.assert_allclose.html) to assert float equality
- Your test cases are an ideal use-case for parameterization (see https://docs.pytest.org/en/latest/parametrize.html). You should be able to collapse the whole test into two arrays (input + expected output).

trojanjay:
@chohner Thank you for the feedback.

I will review and revert accordingly!

chohner:
Cool, let me know if you have questions!
Also no need to revert, simply continue to push to your branch. (Ideally, next time open a new branch on your repository before working on something new so your master can stay untouched).

mgeier:
Closing in favor of #61.

Issue #64, opened 2018-10-24
Example in documentation of sfs.mono.source.point_velocity shows sound pressure of a dipole


spors:
The example in documentation of `sfs.mono.source.point_velocity` shows sound pressure of a dipole. See http://python.sfstoolbox.org/en/0.4.0/frequency-domain.html#sfs.mono.source.point_velocity

mgeier:
The docs are somehow messed up on RTD. The functions were in alphabetical order, even though they should be in the order from the source file.

I've "wiped" and re-built the docs for `master` and `0.4.0` on RTD. Now the order of functions is correct, but the plot images are completely mixed up ... when I build it locally, it looks fine.

hagenw:
If I look at http://python.sfstoolbox.org/en/0.4.0/frequency-domain.html#sfs.mono.source.point_velocity the figures are not mixed up, or do I miss something?

mgeier:
Indeed, now it looks OK! This may have been a caching issue on my browser (or on the server?).

AFAICT, now everything looks as expected, I think we can close this issue.

PR #69, opened 2018-12-27
Pulsating sphere

narahahn:
Simulation of the particle displacement and particle velocity of a pulsating sphere.

pulsating_sphere_d
pulsating_sphere_u

### DIFF
--- a/doc/examples.rst
+++ b/doc/examples.rst
@@ -15,4 +15,5 @@ Examples
 
     examples/modal-room-acoustics
     examples/mirror-image-source-model
+    examples/animations-pulsating-sphere
     example-python-scripts
--- a/doc/examples/animations-pulsating-sphere.ipynb
+++ b/doc/examples/animations-pulsating-sphere.ipynb
@@ -1,0 +1,366 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Animations of a Pulsating Sphere"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import sfs\n",
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "from IPython.display import HTML"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this example, the sound field of a pulsating sphere is visualized.\n",
+    "Different acoustic variables, such as sound pressure,\n",
+    "particle velocity, and particle displacement, are simulated.\n",
+    "The first two quantities are computed with\n",
+    "\n",
+    "- [sfs.mono.source.pulsating_sphere()](../sfs.mono.source.rst#sfs.mono.source.pulsating_sphere) and \n",
+    "- [sfs.mono.source.pulsating_sphere_velocity()](../sfs.mono.source.rst#sfs.mono.source.pulsating_sphere_velocity)\n",
+    "\n",
+    "while the last one can be obtained by using\n",
+    "\n",
+    "- [sfs.util.displacement()](../sfs.util.rst#sfs.util.displacement)\n",
+    "\n",
+    "which converts the particle velocity into displacement.\n",
+    "\n",
+    "A couple of additional functions are implemented in\n",
+    "\n",
+    "- [animations_pulsating_sphere.py](animations_pulsating_sphere.py)\n",
+    "\n",
+    "in order to help creating animating pictures, which is fun!"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import animations_pulsating_sphere as animation"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Pulsating sphere\n",
+    "center = [0, 0, 0]\n",
+    "radius = 0.25\n",
+    "amplitude = 0.05\n",
+    "f = 1000  # frequency\n",
+    "omega = 2 * np.pi * f  # angular frequency\n",
+    "\n",
+    "# Axis limits\n",
+    "figsize = (6, 6)\n",
+    "xmin, xmax = -1, 1\n",
+    "ymin, ymax = -1, 1\n",
+    "\n",
+    "# Animations\n",
+    "frames = 20  # frames per period"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Particle Displacement"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "grid = sfs.util.xyz_grid([xmin, xmax], [ymin, ymax], 0, spacing=0.025)\n",
+    "ani = animation.particle_displacement(\n",
+    "        omega, center, radius, amplitude, grid, frames, figsize, c='Gray')\n",
+    "plt.close()\n",
+    "HTML(ani.to_html5_video())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You can play with the animation more interactively by using `.to_jshtml`."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "HTML(ani.to_jshtml())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Of course, different types of grid can be chosen.\n",
+    "Below is the particle animation using the same parameters\n",
+    "but with a [hexagonal grid](https://www.redblobgames.com/grids/hexagons/)."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def hex_grid(xlim, ylim, hex_edge, align='horizontal'):\n",
+    "    if align is 'vertical':\n",
+    "        umin, umax = ylim\n",
+    "        vmin, vmax = xlim\n",
+    "    else:\n",
+    "        umin, umax = xlim\n",
+    "        vmin, vmax = ylim\n",
+    "    du = np.sqrt(3) * hex_edge\n",
+    "    dv = 1.5 * hex_edge\n",
+    "    num_u = int((umax - umin) / du)\n",
+    "    num_v = int((vmax - vmin) / dv)\n",
+    "    u, v = np.meshgrid(np.linspace(umin, umax, num_u),\n",
+    "                       np.linspace(vmin, vmax, num_v))\n",
+    "    u[::2] += 0.5 * du\n",
+    "\n",
+    "    if align is 'vertical':\n",
+    "        grid = v, u, 0\n",
+    "    elif align is 'horizontal':\n",
+    "        grid = u, v, 0\n",
+    "    return  grid"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "grid = hex_grid([xmin, xmax], [ymin, ymax], 0.0125, 'vertical')\n",
+    "ani = animation.particle_displacement(\n",
+    "        omega, center, radius, amplitude, grid, frames, figsize, c='Gray')\n",
+    "plt.close()\n",
+    "HTML(ani.to_html5_video())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Another one using a random grid."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "grid = [np.random.uniform(xmin, xmax, 4000),\n",
+    "        np.random.uniform(ymin, ymax, 4000), 0]\n",
+    "ani = animation.particle_displacement(\n",
+    "        omega, center, radius, amplitude, grid, frames, figsize, c='Gray')\n",
+    "plt.close()\n",
+    "HTML(ani.to_html5_video())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Each grid has its strengths and weaknesses. Please refer to the\n",
+    "[on-line discussion](https://github.com/sfstoolbox/sfs-python/pull/69#issuecomment-468405536)."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Particle Velocity"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "amplitude = 1e-3\n",
+    "grid = sfs.util.xyz_grid([xmin, xmax], [ymin, ymax], 0, spacing=0.04)\n",
+    "ani = animation.particle_velocity(\n",
+    "        omega, center, radius, amplitude, grid, frames, figsize)\n",
+    "plt.close()\n",
+    "HTML(ani.to_html5_video())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Please notice that the amplitude of the pulsating motion is adjusted\n",
+    "so that the arrows are neither too short nor too long.\n",
+    "This kind of compromise is inevitable since\n",
+    "\n",
+    "$$\n",
+    "\\text{(particle velocity)} = \\text{i} \\omega \\times (\\text{amplitude}),\n",
+    "$$\n",
+    "\n",
+    "thus the absolute value of particle velocity is usually\n",
+    "much larger than that of amplitude.\n",
+    "It should be also kept in mind that the hole in the middle\n",
+    "does not visualizes the exact motion of the pulsating sphere.\n",
+    "According to the above equation, the actual amplitude should be\n",
+    "much smaller than the arrow lengths.\n",
+    "The changing rate of its size is also two times higher than the original frequency."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Sound Pressure"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "amplitude = 0.05\n",
+    "impedance_pw = sfs.default.rho0 * sfs.default.c\n",
+    "max_pressure = omega * impedance_pw * amplitude\n",
+    "\n",
+    "grid = sfs.util.xyz_grid([xmin, xmax], [ymin, ymax], 0, spacing=0.005)\n",
+    "ani = animation.sound_pressure(\n",
+    "        omega, center, radius, amplitude, grid, frames, pulsate=True,\n",
+    "        figsize=figsize, vmin=-max_pressure, vmax=max_pressure)\n",
+    "plt.close()\n",
+    "HTML(ani.to_html5_video())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Notice that the sound pressure exceeds\n",
+    "the atmospheric pressure ($\\approx 10^5$ Pa), which of course makes no sense.\n",
+    "This is due to the large amplitude (50 mm) of the pulsating motion.\n",
+    "It was chosen to better visualize the particle movements\n",
+    "in the earlier animations.\n",
+    "\n",
+    "For 1 kHz, the amplitude corresponding to a moderate sound pressure,\n",
+    "let say 1 Pa, is in the order of micrometer.\n",
+    "As it is very small compared to the corresponding wavelength (0.343 m),\n",
+    "the movement of the particles and the spatial structure of the sound field\n",
+    "cannot be observed simultaneously.\n",
+    "Furthermore, at high frequencies, the sound pressure\n",
+    "for a given particle displacement scales with the frequency.\n",
+    "The smaller wavelength (higher frequency) we choose,\n",
+    "it is more likely to end up with a prohibitively high sound pressure.\n",
+    "\n",
+    "In the following examples, the amplitude is set to a realistic value 1 $\\mu$m.\n",
+    "Notice that the pulsating motion of the sphere is no more visible."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "amplitude = 1e-6\n",
+    "impedance_pw = sfs.default.rho0 * sfs.default.c\n",
+    "max_pressure = omega * impedance_pw * amplitude\n",
+    "\n",
+    "grid = sfs.util.xyz_grid([xmin, xmax], [ymin, ymax], 0, spacing=0.005)\n",
+    "ani = animation.sound_pressure(\n",
+    "        omega, center, radius, amplitude, grid, frames, pulsate=True,\n",
+    "        figsize=figsize, vmin=-max_pressure, vmax=max_pressure)\n",
+    "plt.close()\n",
+    "HTML(ani.to_html5_video())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let's zoom in closer to the boundary of the sphere."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "L = 10 * amplitude\n",
+    "xmin_zoom, xmax_zoom = radius - L, radius + L\n",
+    "ymin_zoom, ymax_zoom = -L, L"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "grid = sfs.util.xyz_grid([xmin_zoom, xmax_zoom], [ymin_zoom, ymax_zoom], 0, spacing=L / 100)\n",
+    "ani = animation.sound_pressure(\n",
+    "        omega, center, radius, amplitude, grid, frames, pulsate=True,\n",
+    "        figsize=figsize, vmin=-max_pressure, vmax=max_pressure)\n",
+    "plt.close()\n",
+    "HTML(ani.to_html5_video())"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This shows how the vibrating motion of the sphere (left half)\n",
+    "changes the sound pressure of the surrounding air (right half).\n",
+    "Notice that the sound pressure increases/decreases (more red/blue)\n",
+    "when the surface accelerates/decelerates."
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python [default]",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.5.6"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 2
+}
--- a/doc/examples/animations_pulsating_sphere.py
+++ b/doc/examples/animations_pulsating_sphere.py
@@ -1,0 +1,114 @@
+"""Animations of pulsating sphere."""
+import sfs
+import numpy as np
+from matplotlib import pyplot as plt
+from matplotlib import animation
+
+
+def particle_displacement(omega, center, radius, amplitude, grid, frames,
+                          figsize=(8, 8), interval=80, blit=True, **kwargs):
+    """Generate sound particle animation."""
+    velocity = sfs.mono.source.pulsating_sphere_velocity(
+               omega, center, radius, amplitude, grid)
+    displacement = sfs.util.displacement(velocity, omega)
+    phasor = np.exp(1j * 2 * np.pi / frames)
+
+    fig, ax = plt.subplots(figsize=figsize)
+    ax.axis([grid[0].min(), grid[0].max(), grid[1].min(), grid[1].max()])
+    scat = sfs.plot.particles(grid + displacement, **kwargs)
+
+    def update_frame_displacement(i):
+        position = (grid + displacement * phasor**i).apply(np.real)
+        position = np.column_stack([position[0].flatten(),
+                                    position[1].flatten()])
+        scat.set_offsets(position)
+        return [scat]
+
+    return animation.FuncAnimation(
+            fig, update_frame_displacement, frames,
+            interval=interval, blit=blit)
+
+
+def particle_velocity(omega, center, radius, amplitude, grid, frames,
+                      figsize=(8, 8), interval=80, blit=True, **kwargs):
+    """Generate particle velocity animation."""
+    velocity = sfs.mono.source.pulsating_sphere_velocity(
+               omega, center, radius, amplitude, grid)
+    phasor = np.exp(1j * 2 * np.pi / frames)
+
+    fig, ax = plt.subplots(figsize=figsize)
+    ax.axis([grid[0].min(), grid[0].max(), grid[1].min(), grid[1].max()])
+    quiv = sfs.plot.vectors(
+            velocity, grid, clim=[-omega * amplitude, omega * amplitude],
+            **kwargs)
+
+    def update_frame_velocity(i):
+        quiv.set_UVC(*(velocity[:2] * phasor**i).apply(np.real))
+        return [quiv]
+
+    return animation.FuncAnimation(
+            fig, update_frame_velocity, frames, interval=interval, blit=True)
+
+
+def sound_pressure(omega, center, radius, amplitude, grid, frames,
+                   pulsate=False, figsize=(8, 8), interval=80, blit=True,
+                   **kwargs):
+    """Generate sound pressure animation."""
+    pressure = sfs.mono.source.pulsating_sphere(
+            omega, center, radius, amplitude, grid, inside=pulsate)
+    phasor = np.exp(1j * 2 * np.pi / frames)
+
+    fig, ax = plt.subplots(figsize=figsize)
+    im = sfs.plot.soundfield(np.real(pressure), grid, **kwargs)
+    ax.axis([grid[0].min(), grid[0].max(), grid[1].min(), grid[1].max()])
+
+    def update_frame_pressure(i):
+        distance = np.linalg.norm(grid)
+        p = pressure * phasor**i
+        if pulsate:
+            p[distance <= radius + amplitude * np.real(phasor**i)] = np.nan
+        im.set_array(np.real(p))
+        return [im]
+
+    return animation.FuncAnimation(
+            fig, update_frame_pressure, frames, interval=interval, blit=True)
+
+
+if __name__ == '__main__':
+
+    # Pulsating sphere
+    center = [0, 0, 0]
+    radius = 0.25
+    f = 750  # frequency
+    omega = 2 * np.pi * f  # angular frequency
+
+    # Axis limits
+    xmin, xmax = -1, 1
+    ymin, ymax = -1, 1
+
+    # Animations
+    frames = 20  # frames per period
+
+    # Particle displacement
+    amplitude = 5e-2  # amplitude of the surface displacement
+    grid = sfs.util.xyz_grid([xmin, xmax], [ymin, ymax], 0, spacing=0.025)
+    ani = particle_displacement(
+            omega, center, radius, amplitude, grid, frames, c='Gray')
+    ani.save('pulsating_sphere_displacement.gif', dpi=80, writer='imagemagick')
+
+    # Particle velocity
+    amplitude = 1e-3  # amplitude of the surface displacement
+    grid = sfs.util.xyz_grid([xmin, xmax], [ymin, ymax], 0, spacing=0.04)
+    ani = particle_velocity(
+            omega, center, radius, amplitude, grid, frames)
+    ani.save('pulsating_sphere_velocity.gif', dpi=80, writer='imagemagick')
+
+    # Sound pressure
+    amplitude = 1e-6  # amplitude of the surface displacement
+    impedance_pw = sfs.default.rho0 * sfs.default.c
+    max_pressure = omega * impedance_pw * amplitude
+    grid = sfs.util.xyz_grid([xmin, xmax], [ymin, ymax], 0, spacing=0.005)
+    ani = sound_pressure(
+            omega, center, radius, amplitude, grid, frames, pulsate=True,
+            colorbar=True, vmin=-max_pressure, vmax=max_pressure)
+    ani.save('pulsating_sphere_pressure.gif', dpi=80, writer='imagemagick')
--- a/sfs/mono/source.py
+++ b/sfs/mono/source.py
@@ -707,6 +707,115 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None, rho0=None):
     return util.XyzComponents([i * n for n in n0])
 
 
+def pulsating_sphere(omega, center, radius, amplitude, grid, inside=False,
+                     c=None):
+    """Sound pressure of a pulsating sphere.
+
+    Parameters
+    ---------
+    omega : float
+        Frequency of pulsating sphere
+    center : (3,) array_like
+        Center of sphere.
+    radius : float
+        Radius of sphere.
+    amplitude : float
+        Amplitude of displacement.
+    grid : triple of array_like
+        The grid that is used for the sound field calculations.
+        See `sfs.util.xyz_grid()`.
+    inside : bool, optional
+        As default, `numpy.nan` is returned for inside the sphere.
+        If ``inside=True``, the sound field inside the sphere is extrapolated.
+    c : float, optional
+        Speed of sound.
+
+    Returns
+    -------
+    numpy.ndarray
+        Sound pressure at positions given by *grid*.
+        If ``inside=False``, `numpy.nan` is returned for inside the sphere.
+
+    Examples
+    --------
+
+    .. plot::
+        :context: close-figs
+
+        radius = 0.25
+        amplitude = 1 / (radius * omega * sfs.default.rho0 * sfs.default.c)
+        p = sfs.mono.source.pulsating_sphere(omega, x0, radius, amplitude, grid)
+        sfs.plot.soundfield(p, grid)
+        plt.title("Sound Pressure of a Pulsating Sphere")
+
+    """
+    if c is None:
+        c = default.c
+    k = util.wavenumber(omega, c)
+    center = util.asarray_1d(center)
+    grid = util.as_xyz_components(grid)
+
+    distance = np.linalg.norm(grid - center)
+    theta = np.arctan(1, k * distance)
+    impedance = default.rho0 * c * np.cos(theta) * np.exp(1j * theta)
+    radial_velocity = 1j * omega * amplitude * radius / distance \
+        * np.exp(-1j * k * (distance - radius))
+    if not inside:
+        radial_velocity[distance <= radius] = np.nan
+    return impedance * radial_velocity
+
+
+def pulsating_sphere_velocity(omega, center, radius, amplitude, grid, c=None):
+    """Particle velocity of a pulsating sphere.
+
+    Parameters
+    ---------
+    omega : float
+        Frequency of pulsating sphere
+    center : (3,) array_like
+        Center of sphere.
+    radius : float
+        Radius of sphere.
+    amplitude : float
+        Amplitude of displacement.
+    grid : triple of array_like
+        The grid that is used for the sound field calculations.
+        See `sfs.util.xyz_grid()`.
+    c : float, optional
+        Speed of sound.
+
+    Returns
+    -------
+    `XyzComponents`
+        Particle velocity at positions given by *grid*.
+        `numpy.nan` is returned for inside the sphere.
+
+    Examples
+    --------
+
+    .. plot::
+        :context: close-figs
+
+        v = sfs.mono.source.pulsating_sphere_velocity(omega, x0, radius, amplitude, vgrid)
+        sfs.plot.soundfield(p, grid)
+        sfs.plot.vectors(v, vgrid)
+        plt.title("Sound Pressure and Particle Velocity of a Pulsating Sphere")
+
+    """
+    if c is None:
+        c = default.c
+    k = util.wavenumber(omega, c)
+    grid = util.as_xyz_components(grid)
+
+    center = util.asarray_1d(center)
+    offset = grid - center
+    distance = np.linalg.norm(offset)
+    radial_velocity = 1j * omega * amplitude * radius / distance \
+        * np.exp(-1j * k * (distance - radius))
+    radial_velocity[distance <= radius] = np.nan
+    return util.XyzComponents([radial_velocity * o / distance for o in offset])
+
+
 def _duplicate_zdirection(p, grid):
     """If necessary, duplicate field in z-direction."""
     gridshape = np.broadcast(*grid).shape
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ addons:
   apt:
     packages:
       - pandoc
+      - ffmpeg
 install:
   - pip install .
   - pip install -r tests/requirements.txt
--- a/sfs/plot.py
+++ b/sfs/plot.py
@@ -328,8 +328,8 @@ def level(p, grid, xnorm=None, power=False, cmap=None, vmax=3, vmin=-50,
 
 
 def particles(x, trim=None, ax=None, xlabel='x (m)', ylabel='y (m)',
-              edgecolor='', **kwargs):
-    """Plot particle positions as scatter plot."""
+              edgecolor='', marker='.', s=15, **kwargs):
+    """Plot particle positions as scatter plot"""
     XX, YY = [np.real(c) for c in x[:2]]
 
     if trim is not None:
@@ -342,12 +342,12 @@ def particles(x, trim=None, ax=None, xlabel='x (m)', ylabel='y (m)',
     if ax is None:
         ax = plt.gca()
 
-    ax.scatter(XX, YY, edgecolor=edgecolor, **kwargs)
-
     if xlabel:
         ax.set_xlabel(xlabel)
     if ylabel:
         ax.set_ylabel(ylabel)
+    return ax.scatter(XX, YY, edgecolor=edgecolor, marker=marker, s=s,
+                      **kwargs)
 
 
 def vectors(v, grid, cmap='blacktransparent', headlength=3, headaxislength=2.5,


doc/examples/pulsating-sphere-displacement.py
```diff
@@ -0,0 +1,72 @@
+"""
+Particle displacement of a pulsating sphere.
+"""
+import sfs
+import numpy as np
+from matplotlib import pyplot as plt
+from matplotlib import animation, patches
+
+# Pulsating sphere
+x0 = [0, 0, 0]  # position
```
mgeier: What about calling this `center`?
Or `position`?

doc/examples/pulsating-sphere-displacement.py
```diff
@@ -0,0 +1,72 @@
...
+a = 0.25  # radius
```
mgeier: What about calling this `radius`?

doc/examples/pulsating-sphere-displacement.py
```diff
@@ -0,0 +1,72 @@
...
+x = np.tile(np.linspace(xmin, xmax, N), N)
+y = np.repeat(np.linspace(ymin, ymax, N), N)
+z = 0
+grid = sfs.util.XyzComponents([x, y, z])
+
+# Particle displacement
+d = sfs.mono.source.pulsating_sphere_displacement(omega, x0, a, d_peak, grid)
+
+# Animation
+fig = plt.figure(figsize=(8, 8))
```
mgeier: It would be great to put all the plotting stuff into a function if that's possible.

And make the whole file `import`able.

The concrete definitions could then go into such a section in the end of the file:

```python
if __name__ == '__main__':
    x0 = ...
    a = ...

    ...

    ani = ...
    ani.save(...)
```

Then the file could also be imported and it would be easier to play around with it.

doc/examples/pulsating-sphere-displacement.py
```diff
@@ -0,0 +1,72 @@
+"""
+Particle displacement of a pulsating sphere.
+"""
```
mgeier:
```
$ python3 -m pydocstyle pulsating-sphere-displacement.py 
pulsating-sphere-displacement.py:1 at module level:
        D200: One-line docstring should fit on one line with quotes (found 3)
```

doc/examples/pulsating-sphere-displacement.py
```diff
@@ -0,0 +1,72 @@
...
+    X = grid + np.real(d * phase_shift)
+    line.set_data(X[0], X[1])
+    circ_patch.set_radius(a + d_peak * np.real(phase_shift))
+    text.set_text('{:0.2f} ms'.format(t[i] * 1000))
+    return line,
+
+
+ani = animation.FuncAnimation(fig,
+                              animate,
+                              frames=np.arange(L),
```
mgeier: frames=L,

doc/examples/pulsating-sphere-displacement.py
```diff
@@ -0,0 +1,72 @@
...
+ax.axis('off')
+
+
+def animate(i, d_peak, line, circ_patch, text):
+    phase_shift = np.exp(1j * omega * t[i])
+    X = grid + np.real(d * phase_shift)
+    line.set_data(X[0], X[1])
+    circ_patch.set_radius(a + d_peak * np.real(phase_shift))
+    text.set_text('{:0.2f} ms'.format(t[i] * 1000))
+    return line,
```
mgeier: If you return something (which is only necessary if `blit` is used), you are supposed to return all modified elements:

https://matplotlib.org/api/_as_gen/matplotlib.animation.FuncAnimation.html

Since you are not using `blit` below, you don't need to return anything.
But if you make the `animate()` function `import`able, it would be nice to return all those things.

sfs/mono/source.py
```diff
@@ -705,6 +705,101 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
     return util.XyzComponents([i * n for n in n0])
 
 
+def pulsating_sphere_displacement(omega, x0, a, d, grid, c=None):
+    """Particle displacement of a pulsating sphere.
+
+    Paramters
```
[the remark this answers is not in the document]
mgeier: You should try to create the docs locally, then this error will be reported.

sfs/mono/source.py
```diff
@@ -705,6 +705,101 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+    grid : triple of array_like
+        The grid that is used for the sound field calculations.
+        See `sfs.util.xyz_grid()`.
+    c : float, optional
+        Speed of sound.
+
+    Returns
+    -------
+    `XyzComponents`
+        Sound pressure at positions given by *grid*.
```
mgeier: It would be great to have an example plot in the docstring!

doc/examples/pulsating-sphere-displacement.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+ani = animation.FuncAnimation(fig,
+                              animate,
+                              frames=np.arange(L),
+                              fargs=(d_peak, line, circ_patch, text))
+ani.save('pulsating_sphere.gif', fps=10, dpi=80, writer='imagemagick')
```
mgeier: I'm getting this warning here:

```
/usr/lib/python3/dist-packages/numpy/core/numeric.py:538: ComplexWarning: Casting complex values to real discards the imaginary part
  return array(a, dtype, copy=False, order=order)
```

I didn't look into it, but you should probably discard the imaginary part somewhere?

sfs/mono/source.py
```diff
@@ -705,6 +705,101 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+    """
+    k = util.wavenumber(omega, c)
+    x0 = util.asarray_1d(x0)
+    grid = util.as_xyz_components(grid)
+
+    r = np.linalg.norm(grid - x0)
+    r[r <= a] = np.nan
+    theta = np.arctan(1, k * r)
+    Z = defs.rho0 * defs.c * np.cos(theta) * np.exp(1j * theta)
+    return Z * 1j * omega * d * a / r * np.exp(-1j * k * (r - a))
```
mgeier: The return type is not a `XyzComponents`?

sfs/mono/source.py
```diff
@@ -705,6 +705,101 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+    `XyzComponents`
+        Particle displacement at positions given by *grid*.
+    """
+    k = util.wavenumber(omega, c)
+    x0 = util.asarray_1d(x0)
+    offset = grid - x0
+    r = np.linalg.norm(offset)
+    r[r <= a] = np.nan
+    return util.XyzComponents([d * a / r**2 * np.exp(-1j * k * (r - a)) * o
+                               for o in offset])
```
mgeier: `XyzComponents` has an `.apply()` method that you could probably use here: http://python.sfstoolbox.org/en/latest/utilities.html#sfs.util.XyzComponents.apply

sfs/mono/source.py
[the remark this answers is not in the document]
narahahn: Oops, thank you for the nice tip.

sfs/mono/source.py
```diff
@@ -705,6 +705,101 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+
+    Returns
+    -------
+    `XyzComponents`
+        Particle velocity at positions given by *grid*.
+    """
+    return 1j * omega * pulsating_sphere_displacement(omega, x0, a, d, grid, c)
+
+
+def pulsating_sphere(omega, x0, a, d, grid, c=None):
```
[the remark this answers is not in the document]
narahahn: Those functions are reordered, and `pulsating_sphere_displacement` is now removed.

sfs/mono/source.py
```diff
@@ -705,6 +705,101 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+    """Sound pressure of a pulsating sphere.
+
+    Paramters
+    ---------
+    omega : float
+        Frequency of pulsating sphere
+    x0 : (3,) array_like
+        Position of source.
+    a : float
```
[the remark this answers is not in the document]
narahahn: Yup, that's much better. `a` and `d` are both changed.

sfs/mono/source.py
```diff
@@ -705,6 +705,101 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+        Radius of sphere.
+    d : float
```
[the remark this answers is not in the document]
narahahn: Thank you!!

sfs/mono/source.py
```diff
@@ -705,6 +705,101 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+        The grid that is used for the sound field calculations.
+        See `sfs.util.xyz_grid()`.
+    c : float, optional
+        Speed of sound.
+
+    Returns
+    -------
+    `XyzComponents`
+        Sound pressure at positions given by *grid*.
+    """
```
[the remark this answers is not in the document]
narahahn: Added both in `pulsating_sphere()` and `pulsating_sphere_velocity()`

sfs/mono/source.py
```diff
@@ -705,6 +705,101 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+    k = util.wavenumber(omega, c)
+    x0 = util.asarray_1d(x0)
+    grid = util.as_xyz_components(grid)
+
+    r = np.linalg.norm(grid - x0)
+    r[r <= a] = np.nan
+    theta = np.arctan(1, k * r)
+    Z = defs.rho0 * defs.c * np.cos(theta) * np.exp(1j * theta)
```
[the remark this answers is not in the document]
narahahn: Oh, my mistake. My [recent commit](0c986db) only fixes for `c`.

sfs/mono/source.py
```diff
@@ -705,6 +705,101 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+
+def pulsating_sphere(omega, x0, a, d, grid, c=None):
+    """Sound pressure of a pulsating sphere.
+
+    Paramters
+    ---------
+    omega : float
+        Frequency of pulsating sphere
+    x0 : (3,) array_like
+        Position of source.
```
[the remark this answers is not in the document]
narahahn: -> "Center of sphere"

doc/examples/pulsating-sphere-displacement.py
```diff
@@ -0,0 +1,72 @@
...
+# Temporal sampling for animation
+fs = f * 10  # sampling frequency
+L = int(np.round(fs / f))  # number of frames corresponding to one period
+t = np.arange(L) / fs  # time
+
+# Uniform grid
+N = 120
+xmin, xmax = -1, 1
+ymin, ymax = -1, 1
+x = np.tile(np.linspace(xmin, xmax, N), N)
```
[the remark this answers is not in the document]
narahahn: Just forgot that nice function.

doc/examples/pulsating-sphere-displacement.py
```diff
@@ -0,0 +1,72 @@
...
+                                         radius=a,
+                                         facecolor='RoyalBlue',
+                                         linestyle='None'))
+line, = ax.plot(grid[0], grid[1],
+                marker='o',
+                markersize=1.5,
+                markerfacecolor='gray',
+                markeredgecolor='None',
+                linestyle='None',
+                alpha=0.75)
```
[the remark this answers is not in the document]
narahahn: I used `pyplot.plot` because it returns a `line` object which makes it easier to update within `animation.FuncAnimation()`. If I get it right, `sfs.plot.particles` isn't returning anything.

doc/examples/pulsating-sphere-displacement.py
narahahn: Other example files are using similar names like `x0` or `xs`.
Should we rename those too?

doc/examples/pulsating-sphere-displacement.py
narahahn: Thanks!!

doc/examples/pulsating-sphere-displacement.py
narahahn: This one is fixed, thank you.

sfs/mono/source.py
narahahn: Scalar fields should be `numpy.ndarray` not `XyzComponents`, right?

sfs/mono/source.py
narahahn: That's a very useful feature.
But how can I use it in this case? Should I define a new function? Or is there a more pythonic way?

doc/examples/pulsating-sphere-displacement.py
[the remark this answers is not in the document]
mgeier: Feel free to change `sfs.plot.particles()` to return whatever makes sense!

doc/examples/pulsating-sphere-displacement.py
mgeier: I don't know. Probably.

I just find it absurd to call something `x0` and add a comment saying "position".
If `x0` is already communicating the fact that it is a position, you should remove the comment. If not, you should probably rename the variable (and remove the comment as well).

sfs/mono/source.py
mgeier: Yes, of course you are right!

I was just confused because the resulting array was 1-dimensional, but that was just because your `grid` was broken (which is fixed in the meantime) and because the docstring was claiming that a `XyzComponents` is returned.

The same mistake was made in the `point()` and `plane()` functions, see #72.

sfs/mono/source.py
mgeier: I'm not sure if it's really worth to use `.apply()` in this case ...

But if you do, you could use a `lambda` function instead of defining a separate function before.
However, it is uncommon to have a line break in a `lambda` function. Line breaks in list comprehensions are more common.

Probably you should just leave it as it is ...?

mgeier:
Thanks for adding 2 more animations, but wouldn't it be better to make a single file that generates all 3 animations? Right now there is a lot of repetition.

sfs/mono/source.py
```diff
@@ -705,6 +705,80 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+    k = util.wavenumber(omega, c)
+    x0 = util.asarray_1d(x0)
+    grid = util.as_xyz_components(grid)
+
+    r = np.linalg.norm(grid - x0)
+    r[r <= radius] = np.nan
+    theta = np.arctan(1, k * r)
+    impedence = defs.rho0 * c * np.cos(theta) * np.exp(1j * theta)
+    return impedence * 1j * omega * amplitude * radius / r \
+        * np.exp(-1j * k * (r - radius))
```
mgeier: I'm getting the warning

    /home/mg/git/sfs-python/sfs/mono/source.py:743: RuntimeWarning: invalid value encountered in true_divide
      * np.exp(-1j * k * (r - radius))

I guess this is because of the NaN values you are using.

Is it possible to add the NaN values after the calculation?
Or excluding them from the calculation somehow?

You could also silence the warning, but this should only be the last resort if nothing else works.

sfs/mono/source.py
```diff
@@ -705,6 +705,80 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+    if c is None:
+        c = defs.c
+    k = util.wavenumber(omega, c)
+    x0 = util.asarray_1d(x0)
+    offset = grid - x0
+    r = np.linalg.norm(offset)
+    r[r <= radius] = np.nan
+    return util.XyzComponents([1j * omega * amplitude * radius / r**2
+                               * np.exp(-1j * k * (r - radius)) * o
+                               for o in offset])
```
mgeier: /home/mg/git/sfs-python/sfs/mono/source.py:779: RuntimeWarning: invalid value encountered in true_divide
      for o in offset])

See above.

doc/examples/pulsating-sphere-displacement.py
narahahn: No it wasn't, sorry.
I wasn't aware of `pydocstyle` and have only checked
```
$ pycodestyle
```
Thanks.

doc/examples/pulsating-sphere-displacement.py
narahahn: I got your point. You're right.

sfs/mono/source.py
narahahn: I also noticed that after your comment.
Actually I've copy-and-pasted from the `point()` function. I should have checked it.

sfs/mono/source.py
narahahn: I would keep it as it is, but please let me know if you come up with a better idea.

doc/examples/animation-pulsating-sphere-pressure.py
```diff
@@ -0,0 +1,77 @@
...
+
+# Sound pressure
+pressure = sfs.mono.source.pulsating_sphere(omega,
+                                            x0,
+                                            radius,
+                                            amplitude,
+                                            grid)
+
+# Animation
+impedence_pw = sfs.defs.rho0 * sfs.defs.c
```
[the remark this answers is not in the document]
narahahn: Aarg!

sfs/mono/source.py
```diff
@@ -705,6 +705,80 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+    grid : triple of array_like
+        The grid that is used for the sound field calculations.
+        See `sfs.util.xyz_grid()`.
+    c : float, optional
+        Speed of sound.
+
+    Returns
+    -------
+    numpy.ndarray
+        Sound pressure at positions given by *grid*.
```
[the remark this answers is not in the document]
narahahn: Missing line is now added.

narahahn:
The commit message for 4421b3c is supposed to be

> Use `sfs.plot.particles` for displacement animation.

sfs/mono/source.py
mgeier: OK.

sfs/mono/source.py
mgeier: Yes, thanks, the warning is gone now!

I don't think I did anything to switch this warning on. At least I cannot remember.

I'm using NumPy version `1.16.0rc1` from a Debian package.

doc/examples/animation-pulsating-sphere-displacement.py
```diff
@@ -0,0 +1,74 @@
...
+ax = plt.axes(xlim=(-xmax, xmax), ylim=(-ymax, ymax))
+patch = ax.add_patch(patches.Circle(center[:2],
+                                    radius=radius,
+                                    facecolor='RoyalBlue',
+                                    linestyle='None',
+                                    alpha=0.4))
+scat = sfs.plot.particles(grid + displacement,
+                          s=15,
+                          c='gray',
+                          marker='.')
```
mgeier: What about changing the default values of `sfs.plot.particles()` to be more sensible?

I think the only other use is in the `plot_particle_density.py` example.

narahahn:
@mgeier Thank's a lot for the review and comments. They are very helpful.
Your suggestions are considered in the recent commit 40dfb37.

BTW, how can I update the docs without re-building all of them?
Is it possible to build only a selected part?

sfs/mono/source.py
```diff
@@ -21,6 +22,12 @@
     # Grid for vector fields:
     vgrid = sfs.util.xyz_grid([-2, 3], [-1, 2], 0, spacing=0.1)
 
+    # Parameters for pulsating sphere:
```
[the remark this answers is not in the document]
narahahn: Yes, I also think so.

doc/examples/animation-pulsating-sphere-displacement.py
narahahn: I think I don't understand what you mean by `more sensible`.
Do you mean that the particle displacement should be amplified by default, so that it is better visualized even if you use realistic values which in the order of micrometers?

sfs/mono/source.py
```diff
@@ -705,6 +705,99 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+    Examples
+    --------
+
+    .. plot::
+        :context: close-figs
+
+        radius = 0.25
+        amplitude = 1 / (radius * omega * sfs.defs.rho0 * sfs.defs.c)
+        p = sfs.mono.source.pulsating_sphere(omega, x0, radius, amplitude, grid)
+        v = sfs.mono.source.pulsating_sphere_velocity(omega, x0, radius, amplitude, vgrid)
```
mgeier: This line has more than 79 characters (more specifically it has 82) when shown in the docs.

doc/examples/animation-pulsating-sphere-displacement.py
mgeier: No, i don't mean the amplification. I think this should be done explicitly every single time it is used.

I meant the `marker`, `s` and `c` arguments. Because the default sizes are definitely too large for our use case, and the color black seems to be a reasonable default in our case, right?

I have never thought about using `plot()`, because in my mind this means plotting some kind of lines between connected dots, but according to the docs it does "Plot y versus x as lines and/or markers", so "lines" are not required.

I don't foresee that anyone will want to use different markers in the same plot.
But I could definitely imagine using a color gradient, e.g. for the (scalar) speed of the particles or probably for the pressure.
It might even make sense to change the size of the markers, e.g. according to sound pressure, but I didn't try this. Probably it doesn't make sense ...

I think `scatter()` is a better fit, but if you prefer `plot()`, I'm also fine with this.

mgeier:
> BTW, how can I update the docs without re-building all of them?
> Is it possible to build only a selected part?

By default, Sphinx should only build the pages with changes (unless you use the `-E` or `-a` option).
Does it not do that for you?

For further options see

    $ python3 -m sphinx --help

doc/examples/animation-pulsating-sphere-displacement.py
narahahn: Yes, the default size is indeed too large. I would also prefer circles with black `facecolor`. I find it also nice to give some transparency (`alpha`) as this better visualizes the particle density. Should I change the default values of `sfs.plot.particles` in this PR?

I also think we can keep using `scatter`. Although the documentation says that `plot` is faster, it wouldn't be that significant in our current use cases.

By the way, visualizing the particle velocity with colors sounds nice. I'll make an example.

narahahn:
Particle animation with colors.

pulsating_sphere_displacement

The radial component of the particle velocity is indicated by color gradient. Red means positive radial velocity (outward movement) whereas blue means negative radial velocity (inward movement).

The red wavefronts (positive) are a bit narrower than the blue ones (negative). This is because the particles are compressed when the particles are moving outward. When the particles are moving inward, on the other hand, the particles are rarefied and have a wider spatial extent.

doc/examples/animation-pulsating-sphere-displacement.py
mgeier: > Should I change the default values of `sfs.plot.particles` in this PR?

Sure, why not!
But if you prefer, you can of course also make a new PR.

hagenw:
I would say it is not only an optical illusion and I would suspect that near the sphere it also has to be that way. Maybe it will look more equally distributed if you go farer away from the source?

narahahn:
I think the reason for the "illusion" is the movement of particles. The velocity is computed for a uniform grid, which is the equilibrium points of the particles. Then the corresponding colors are applied to a nonuniform grid, because the particles are swinging around the equilibrium. The resulting plot is therefore "distorted". If the colors were plotted at the equilibrium, there would be no such illusion.

Another thing to notice is that the amplitude of the displacement is set to 5 cm which is way too large. For 750 Hz, this corresponds to about 24750 Pa at 1 m distance, which makes no sense. To have about 1 Pa, the amplitude should be in the order of millimeters. The animation then look like this:

pulsating_sphere_displacement

The particles are at rest (probably moving in the order of subpixel) which is kinda boring. But now the "rings" have the same widths.

For educational purposes, I find the earlier (insane) animation more useful as it (even without the colors) visualizes the particle velocity and density (therefore the pressure) at the same time. Maybe the colored animation, although it looks nice, should be used/shown with care.

> > The text in the docstring is updated but the plot is not.
> 
> This sounds a bit like #64. This is probably because the old images are cached by the browser?

Thank you, I will check.

mgeier:
> If the colors were plotted at the equilibrium, there would be no such illusion.

But this would be physically wrong, wouldn't it?

It makes sense to show the speed (by means of the color) of the particle at the point in space where the particle actually is at the given point in time (or rather at a given phase angle). At the maximum displacement, the velocity should be 0. But it wouldn't be correct to show this speed at the equilibrium point, because the particle isn't there in the moment. And a different particle which happens to be there in that moment will actually have a different speed!

I now think that this is actually a physical effect, the rings of particles with negative speed should be slightly wider than those with positive speed.

Of course this is all extremely exaggerated, but I think without exaggeration it's useless.

I think it would be best to show a plot with the actual (unnoticeable) displacement and directly afterwards show the exaggerated plot. This would be similar to what I did with the point and line sources (see http://python.sfstoolbox.org/en/0.4.0/frequency-domain.html#sfs.mono.source.point), where I first showed the plot where hardly anything can be seen, and afterwards I showed the normalized plot.

I've seen particle plots where a single particle (or a few selected ones) was plotted in a different color (and probably slightly bigger?), this made it much easier to see the actual displacement of a single particle.

BTW, @spors used a randomized grid in https://github.com/sfstoolbox/sfs-python/blob/master/doc/examples/plot_particle_density.py. I think this looks better and the patterns of the regular grid can be distracting.

narahahn:
> I now think that this is actually a physical effect, the rings of particles with negative speed should be slightly wider than those with positive speed.

I agree that this would actually happens for oscillating particles and the velocity should be visualized at the corresponding position, no matter where it is. What I am not sure about is whether we can use the wave propagation model for arbitrarily large particle displacement.

Let me approach this from a slightly different point of view. I'm not so sure about this, though.
The deformation of the waveform is sort of a nonlinear effect. Wave propagations do exhibit nonlinearities for high amplitudes but in a different manner. If I've understood it correctly (somewhere in the Kuttruff's book), the wave propagates faster than c (speed of sound) for very large positive sound pressure and slower than c for very small negative sound pressure. As a result, a sine wave becomes like a sawtooth wave.

In our simulation, we are assuming constant speed of sound and linear acoustics. If we want to show something that really happens for large amplitude, the variation of the propagation speed has to be taken into account. In my opinion, the nonlinear effect we are observing is sort of an artifact caused by applying linear acoustics to nonlinear range.

> Of course this is all extremely exaggerated, but I think without exaggeration it's useless.

> I think it would be best to show a plot with the actual (unnoticeable) displacement and directly afterwards show the exaggerated plot. This would be similar to what I did with the point and line sources (see http://python.sfstoolbox.org/en/0.4.0/frequency-domain.html#sfs.mono.source.point), where I first showed the plot where hardly anything can be seen, and afterwards I showed the normalized plot.

That's true. I'm not saying that we should always stick to realistic values. As you mentioned, we often normalize the sound pressure for a better visualization. The particle displacement should be amplified for the same reason. But adding colors to such an exaggerated simulation shows a sound field with a modified spatial structure. This might lead innocent readers/users to misunderstanding. Maybe I am being unnecessarily cautious:-)?

> I've seen particle plots where a single particle (or a few selected ones) was plotted in a different color (and probably slightly bigger?), this made it much easier to see the actual displacement of a single particle.

Right, I also saw it before. Wouldn't be a problem. I will make new animation.

> 
> BTW, @spors used a randomized grid in https://github.com/sfstoolbox/sfs-python/blob/master/doc/examples/plot_particle_density.py. I think this looks better and the patterns of the regular grid can be distracting.

IMHO, the uniform grid is a better representation of the particle.
In Kinsler (Ch. 5, p. 114),

> The terms _fluid element_ and _particle_ mean an infinitesimal volume of the fluid large enough to contain millions of molecules so that the fluid may be thought of as a continuous medium, yet small enough that all acoustic variables are uniform throughout.

The particle is not referring to individual air molecule but a chunk of air which can be assumed to be an infinitesimal volume element. Thus, each point in the simulation should be seen as the center of mass of the volume element. If there is no sound wave, all volume elements should be of the same size and equidistant because there is no net flow or force among them. So the grid points must be uniform when there is no sound wave.

Of course, we might want to simulate the air molecules (not the _sound particles_) using random grids. But then the particle velocity shouldn't be used to simulate the individual air molecules, because it is not the air molecules that have such a motion. They would rather have a Brownian motion.
According to Kinsler (ibid),

> The molecules of a fluid do not have fixed mean positions in the medium. Even without the pressure of an acoustic wave, they are in constant random motion with average velocities far in excess of any particle velocity associated with the wave motion.

narahahn:
> The deformation of the waveform is sort of a nonlinear effect. Wave propagations do exhibit nonlinearities for high amplitudes but in a different manner. If I've understood it correctly (somewhere in the Kuttruff's book), the wave propagates faster than c (speed of sound) for very large positive sound pressure and slower than c for very small negative sound pressure. As a result, a sine wave becomes like a sawtooth wave.

This can be found in [Kuttruff, Ch. 4 p. 67] but my explanation was incorrect. The steepening/flattening is observed after a long propagation. Although more pronounced for larger amplitudes, the effect would also occur for not-so-crazy amplitudes. I think this effect is not relevant to our discussion since we are in the near field.

The nonlinear effect that would occur in near field and at high amplitude is the expansion and compression of the waveform for positive and negative amplitudes, respectively. This kind of distortion is illustrated in [ibid, Ch. 2, p. 33] for a exemplary nonlinear system.

Now I am wondering whether we can actually simulate the second nonlinear effect with the current model.

spors:
> Now I am wondering whether we can actually simulate the second nonlinear effect with the current model.

Since the model is based on the linear losless wave-equation, nonlinear effects should not be included.

narahahn:
> I've seen particle plots where a single particle (or a few selected ones) was plotted in a different color (and probably slightly bigger?), this made it much easier to see the actual displacement of a single particle.

Something like this, right?

pulsating_sphere_displacement

narahahn:
> Since the model is based on the linear losless wave-equation, nonlinear effects should not be included.

Yes, you're right. The acoustic variables cannot include any nonlinearities. The conversion among `displacement`, `velocity`, and `pressure` is performed only by linear operations.

But the visualized result of the particles does have some nonlinear effect. The waveform is distorted as the particles moving outward (positive radial direction) are more close together (higher density) than those moving inward (negative radial direction; lower density).

My hypothesis is that this is due to the nonlinear relation of the particle velocity `v` and density `rho`. In linear acoustics, `rho` is assumed to have linear relation with `v` for very small `v` (Kinsler Ch. 5 p. 116; Kuttruff Ch. 3 p. 38). This is one of the building blocks of the linearized wave equation. When `v` is large and cannot be neglected, nonlinear effect will occur.

Here are simulation results for a 1-dimensional plane wave in the x-direction.
velocity_vs_density_microm

* Left-Top: particle displacement
* Right-Top: particle velocity
* Left-Bottom: density vs air density (constant)

The particle displacement is in the order of micrometer, which corresponds to about 1 Pa. The sound pressure is obtained as `v * rho0 * c`. The density is computed numerically, `mass` divided by `volume` for each volume element (particle). As you can see, the variation of the density is very small compared to the air density. The particle displacement (left-top) and particle velocity (right-top) are plotted at their _displaced_ positions, as been done for the pulsating sphere.
```
plt.plot(x + displacement, displacement)
plt.plot(x + displacement, velocity)
```

The results for displacement of millimeter look like this:
velocity_vs_density_mm

The corresponding sound pressure is in the order of 1000 Pa. Although we cannot use the formula `p = rho0 * c * v` for such high pressure, the value is given just for the sake of comparison. The density is changing in a fairly linear manner, i.e. no considerable harmonic distortion.

Finally, the results for displacement of 5 centimeters:
velocity_vs_density_cm

The sound pressure is more than half the atmospheric pressure 10^5 Pa. Now the change of the particle velocity (top-right) is no more simple harmonic, and shows a similar behavior as observed in the pulsating sphere animation. The peaks are narrower than the valleys (dashed line indicates a pure sinusoid). The density also exhibits a strong nonlinearity. The increase of the density is very steep on its maxima, compared to the decrease at minima.

I presume that the particle animation is showing the same kind of nonlinearity caused by large particle velocity (or equivalently particle displacement). Again, this is not about the nonlinear relation of density and pressure, but the the nonlinear relation of particle velocity and density. Converting `v` or `rho` into sound pressure will be a different story.
As I am not super confident of this theory, I would be very happy to hear your thoughts.

mgeier:
>> I've seen particle plots where a single particle (or a few selected ones) was plotted in a different color (and probably slightly bigger?), this made it much easier to see the actual displacement of a single particle.
>
> Something like this, right?

Exactly!

mgeier:
@narahahn Now there are multiple scripts creating the same animations, right?
Can you please remove those that you don't want to keep?

narahahn:
The old scripts are removed.

In `sfs.plot.particles()`, default values `marker='.'` and `s=15` are added.
Those are passed to `scatter()`.

mgeier:
Thanks for the update!

Why is the hole in the pressure plot not changing its size?
It did so in your example above (https://github.com/sfstoolbox/sfs-python/pull/69#issuecomment-450241712).

Thanks for trying to make it `import`able (as I suggested in https://github.com/sfstoolbox/sfs-python/pull/69#discussion_r244495950). Sadly, this doesn't work at all.
The file name is not a valid module name. Many functions access variables that are not defined when `import`ing.

Probably this was too ambitious?
Probably a good old simple script would suffice?

I don't mind either way, but if you decide to make it `import`able, you should actually make it work, and if you decide to make it a simple script, you should simplify it ...

narahahn:
> Why is the hole in the pressure plot not changing its size?
> It did so in your example above ([#69 (comment)](https://github.com/sfstoolbox/sfs-python/pull/69#issuecomment-450241712)).

This is because now the sound pressure is computed only for outside the sphere.
https://github.com/sfstoolbox/sfs-python/blob/9bcb0631a544634a9596b6af50904795a2254743/sfs/mono/source.py#L756

Before this line was added, the sound field inside the sphere was also computed.
Although the earlier animation was physically incorrect,
it did look better in my opinion.

Maybe I can add an option to choose whether the sound field inside the sphere
should be computed or not. Then one can choose between physics and aesthetics.

> I don't mind either way, but if you decide to make it `import`able, you should actually make it work, and if you decide to make it a simple script, you should simplify it ...

Thank you for the review. Let me try to fix it.

doc/examples/animations-pulsating-sphere.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+def init_2d_plot():
+    fig = plt.figure(figsize=(8, 8))
+    ax = plt.axes(xlim=(xmin, xmax), ylim=(ymin, ymax))
+    patch = ax.add_patch(patches.Circle(center[:2],
+                                        radius=radius,
+                                        facecolor='RoyalBlue',
+                                        linestyle='None',
+                                        alpha=0.4))
```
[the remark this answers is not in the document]
narahahn: I was also wondering how to break the lines in such cases.
Thanks!

sfs/mono/source.py
```diff
@@ -705,6 +705,111 @@ def plane_averaged_intensity(omega, x0, n0, grid, c=None):
...
+
+    Examples
+    --------
+
+    .. plot::
+        :context: close-figs
+
+        radius = 0.25
+        amplitude = 1 / (radius * omega * sfs.defs.rho0 * sfs.defs.c)
+        p = sfs.mono.source.pulsating_sphere(omega, x0, radius, amplitude, grid)
```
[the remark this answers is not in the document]
narahahn: Oh, good to know, thanks.

mgeier:
> This is because now the sound pressure is computed only for outside the sphere.
> ...
> Although the earlier animation was physically incorrect,
it did look better in my opinion.

Hmmm, this is unfortunate.

The actual radius depends on the selected phase angle, doesn't it?

Probably you could use the minimum radius?
Then the size of the hole could be adapted in the animation code?

narahahn:
> The actual radius depends on the selected phase angle, doesn't it?

That's right.

> Probably you could use the minimum radius?
> Then the size of the hole could be adapted in the animation code?

How about computing the entire sound field without a hole
and using a non-transparent patch (`alpha=1`) to mask the circular area (see below)?
The hole won't be a perfect circle and will look ugly anyway.

pulsating_sphere_pressure

mgeier:
> How about computing the entire sound field without a hole

Yeah, I guess we'll have to do that.
But you should mention in the docstring that the values inside the sphere are invalid.

And you should correctly apply the hole in the example plot in the docstring, ideally using a non-default phase angle, so that users see how this should be used.

> and using a non-transparent patch

I don't like the whole patch idea anyway because it makes the real data harder to see.

The transparent patch at least doesn't completely hide the data.
I think an unfilled circle would be better to show the sphere if necessary, but most of the time I think the hole should speak for itself.

> The hole won't be a perfect circle and will look ugly anyway.

It may look ugly, but it's the correct data for the given grid values. It doesn't have to be beautiful.

narahahn:
> I don't like the whole patch idea anyway because it makes the real data harder to see.

For sound pressure animation, yes.

> The transparent patch at least doesn't completely hide the data.
> I think an unfilled circle would be better to show the sphere if necessary, but most of the time I think the hole should speak for itself.

I agree. It was just a suggestion for an in-my-opinion-nice-looking animation.

> And you should correctly apply the hole in the example plot in the docstring, ideally using a non-default phase angle, so that users see how this should be used.
> 
> It may look ugly, but it's the correct data for the given grid values. It doesn't have to be beautiful.

If the correctness is of primary interest,
I think the size of the hole shouldn't be changed.
The data is valid only for r >= a.

Maybe we can make the computation inside the sphere optional?
Either a hole with r = a, or no hole.
A hole corresponding to any phase angle will be confusing.

mgeier:
> The data is valid only for r >= a.

I don't understand.
Is it a pulsating sphere with non-changing radius? So it's not pulsating after all?

> A hole corresponding to any phase angle will be confusing.

But there will always be a phase angle, even if it's zero, right?

narahahn:
> I don't understand.
> Is it a pulsating sphere with non-changing radius? So it's not pulsating after all?

I think this contradiction arises because we are using
the pulsating sphere model for large amplitudes
which the model is not suited for.
The amplitude of the pulsating motion, is assumed to be
negligibly small compared to the radius and the wavelength.
The radius is therefore practically constant.

Here is another thing.
From a physical point of view, the sound pressure at the surface
should be maximum when the radius is `r - a` and minimum when `r + a`
(neglecting the phase shift due to the complex impedance).
However, if the current model is used to compute the sound pressure
at `r - a` and `r + a`, it will be apart from the maximum and minimum
by phase shift of `np.exp(+1j * k * a)` and `np.exp(-1j * k * a)`, respectively.

mgeier:
I'm totally confused. But that shouldn't keep us from making progress!

Since the values so far used are extremely exaggerated (because otherwise nothing would be visible at that scale), what about making an additional animation (doesn't have to be all 3 variants) with realistic values?

I tried using this:

```python
center = 0, 0, 0
radius = 0.1
amplitude = 0.0001
xmin, xmax = radius - 0.001, radius + 0.001
ymin, ymax = -0.001, 0.001
grid = sfs.util.xyz_grid([xmin, xmax], [ymin, ymax], 0, spacing=0.00005)
```

I don't know if those are realistic values, but it didn't work anyway. Can you make this or something similar work?

This is a good test for your animation functions anyway, because it shows potential bugs ...

And it shows that plotting the whole sphere-patch in such a general function isn't a good idea.
If you plot the circle, you shouldn't allow an arbitrary `grid` to be passed in.
Or you should trim the circle to the plot limits, which shouldn't be that complicated.

narahahn:
> And it shows that plotting the whole sphere-patch in such a general function isn't a good idea.

Well, in the beginning, it was just intended for making a couple of
animations for the communications acoustics lecture/exercises,
so I tried to add some _decorations_ that might be informative.
General usage wasn't considered.
But now I'd also prefer making it simpler and more versatile.
One can add those kind of things whenever needed.
I removed the patches and texts in the new commit d65a541

> I don't know if those are realistic values, but it didn't work anyway. Can you make this or something similar work?

Thank you for testing it.  It will work now.
The amplitude for a given sound pressure is frequency dependent.
You can roughly check it by
```python
p = 1j * omega * amplitude * rho0 * c
```

> Since the values so far used are extremely exaggerated (because otherwise nothing would be visible at that scale), what about making an additional animation (doesn't have to be all 3 variants) with realistic values?

I set the amplitude for sound pressure and velocity in the order of micrometers,
which gives the sound pressure of around 1 Pa.
But, as we discussed earlier, this wouldn't make any visual differences
in the animations, since the arrow length and colors are normalized anyway,
and the spatial structure won't change.

mgeier:
OK, if realistic amplitudes are in the order of micrometers it won't be possible to do what I initially thought about.

I was hoping we could zoom in far enough to see the movement of the sphere while choosing a high frequency that still allows us to see the structure of the sound field (which should be in the order of centimeters at 20 kHz). This won't work.

But if we zoom to the micrometer level, we could still make an animation, even if the resulting plot is much smaller than a wavelength. We would only see changing colors, but that might still be worth an animation.

What about using something like this (at the end of your example notebook, using some of the previously defined values):

```python
%matplotlib
amplitude = 0.000_001
xmin, xmax = radius - 0.000_005, radius + 0.000_005
ymin, ymax = -0.000_005, 0.000_005
grid = sfs.util.xyz_grid([xmin, xmax], [ymin, ymax], 0, spacing=0.000_01)
ani = animation_sound_pressure(omega, center, radius, amplitude, grid, frames)
```

(the `%matplotlib` thing is to open a plot window for the animations, just for testing)

This would use a realistic frequency (your original 750 Hz) and a realistic amplitude (1 micrometer), leading to realistic sound pressure values.
At this zoom level, the movement of the sphere should be clearly visible.

The only unrealistic thing would be the time scale, because the animation would run much much slower than reality. But that's OK.

Would it make sense to show the changing size of the "hole" in such an example?

Please note that in this example, the sound pressure is not normalized!
The red and blue values are actually reaching magnitudes of about a Pascal.

Since the time scale is not realistic, the used frequency actually doesn't really matter in this case (since we don't see the structure of the sound field anyway), it just has to be chosen appropriately together with the amplitude value to lead to the desired sound pressure.

> [...] since the arrow length and colors are normalized anyway

I think we should avoid the temptation to normalize whenever possible.

I was playing around with the `amplitude` in your example and didn't see any effect on the sound pressure. I thought there was a bug somewhere. It took me quite a while to find out that you were normalizing using `vmin` and `vmax` and therefore my changes of `amplitude` didn't result in any visible changes in the animation ...

Removing the normalization made the thing much easier to understand for me.

> The amplitude for a given sound pressure is frequency dependent.

This is interesting, you could probably mention this in the notebook?

mgeier:
Coming back to the idea of using a randomized grid, which I suggested in https://github.com/sfstoolbox/sfs-python/pull/69#issuecomment-452055202. I said:

> I think this looks better and the patterns of the regular grid can be distracting.

I understand what you (@narahahn) said in https://github.com/sfstoolbox/sfs-python/pull/69#issuecomment-452136422 about the points being the centers of some small volume elements and not the actual particles. I'm absolutely with you on that.

But still, I think a rectangular grid looks bad.

I finally have an example that shows the problem, hopefully quite clearly:

pulsating_sphere_displacement

When contracting, the hole in the middle doesn't look like a circle, it's more like a polygon.
And there are two fun artifacts where at the left of the hole and at the bottom there is a dot missing (I guess due to rounding).
Also, if you look at the propagation of the waves horizontally, this looks quite different than if you look at the propagation along a 45 degree line.
There are slightly different artifacts in the GIF at the very top of this page, which may be subtle, but I find them distracting.
Those structures/artifacts are caused by using a rectangular grid and do not reflect reality.

When using a randomized grid, there might also be some unwanted artifacts, but I guess it is less likely.

I guess some of this is a matter of taste, so I think there is no right or wrong, I'm just giving my very subjective opinion.

narahahn:
> I was hoping we could zoom in far enough to see the movement of the sphere while choosing a high frequency that still allows us to see the structure of the sound field (which should be in the order of centimeters at 20 kHz). This won't work.

Right, you either end up with a very high sound pressure
or an invisibly small particle displacement 

> But if we zoom to the micrometer level, we could still make an animation, even if the resulting plot is much smaller than a wavelength. We would only see changing colors, but that might still be worth an animation.

It sure is.
So it will look like a vibrating plane.

pulsating_sphere_displacement
pulsating_sphere_pressure

> Would it make sense to show the changing size of the "hole" in such an example?

You mean in the sound pressure animation, right? Hmm, interesting point.
As the wavelength is much longer than the amplitude, it won't do any harm.
Moreover, it also visualizes how the pushing/pulling action
creates high/low pressure which is not shown in the particle animation.
Let me try it.

> I was playing around with the `amplitude` in your example and didn't see any effect on the sound pressure. I thought there was a bug somewhere. It took me quite a while to find out that you were normalizing using `vmin` and `vmax` and therefore my changes of `amplitude` didn't result in any visible changes in the animation ...

Sorry for causing confusion.
I did that because `sfs.plot.soundfield` clips the values
and couldn't see the real value in the colorbar.

> > The amplitude for a given sound pressure is frequency dependent.
> 
> This is interesting, you could probably mention this in the notebook?

Yes, sure.
Actually this explains why woofers need
larger amplitudes than tweeters.

narahahn:
The hole in the sound pressure is now optional.
I added an example in the notebook
where the size of the hole constantly changes during the animation.

pulsating_sphere_displacement_zoom
pulsating_sphere_pressure_zoom

narahahn:
> I guess some of this is a matter of taste, so I think there is no right or wrong, I'm just giving my very subjective opinion.

Yeah, it's just a model after all.
It depends what do you want to see or explain.

Here is another reasons why I prefer uniform gird over random gird:
Uniform girds better show the mass-and-spring behavior of the particles,
i.e. the compliance/elasticity of the medium.
Two particles repel when they get closer and attract when move away.
I think this is an important property as it enables the sound wave to propagate.
In a random gird, some particles are very close together but they don't push each other away.
So the particles look like they are just floating on a wavy water.
Of course, this also shows the spatial structure of the wave field,
but not the interaction of the particles.
This is also related to the point I mentioned earlier in this PR.
If there is no sound field, there will be no other forces acting on the particles,
so they should be in their equilibrium points.

pulsating_sphere_displacement_random


> When contracting, the hole in the middle doesn't look like a circle, it's more like a polygon.

This is less noticeable if we add a patch :-)

> And there are two fun artifacts where at the left of the hole and at the bottom there is a dot missing (I guess due to rounding).

Yeah, that's unfortunate.
As you seen in my previous comment,
this also happens even when the radius is very large
compared to the figure. 

> Those structures/artifacts are caused by using a rectangular grid and do not reflect reality.

Let's see how a hexagonal grid works.

narahahn:
Animations with hexagonal grid.

Basically not so different from the square grid.
To my eyes, they have even more artifacts.

1. Vertical layout
pulsating_sphere_displacement_vhex

2. Horizontal layout
pulsating_sphere_displacement_hhex

doc/examples/animations_pulsating_sphere.py
```diff
@@ -0,0 +1,127 @@
+"""Animations of pulsating sphere."""
+import sfs
+import numpy as np
+from matplotlib import pyplot as plt
+from matplotlib import animation
+
+
+def update_frame_displacement(i, grid, displacement, phasor, scat, frames):
```
mgeier: This is only used once, so what about defining it within `animation_particle_displacement()`?

This way you could capture the variables and avoid specifying `fargs`.

doc/examples/animations_pulsating_sphere.py
```diff
@@ -0,0 +1,127 @@
...
+    displacement = sfs.util.displacement(velocity, omega)
+    phasor = np.exp(1j * 2 * np.pi / frames)
+
+    fig = plt.figure(figsize=figsize)
+    ax = plt.axes()
+    scat = sfs.plot.particles(grid + displacement, **kwargs)
+    ax.axis([grid[0].min(), grid[0].max(), grid[1].min(), grid[1].max()])
+    return animation.FuncAnimation(
+            fig, update_frame_displacement, frames,
+            fargs=(grid, displacement, phasor, scat, frames))
```
mgeier: You should probably pass `blit=True` and make it possible to pass in `interval` (but provide a meaningful default).

I guess `repeat` is not necessary, because we always want `repeat=True`, right?

doc/examples/animations_pulsating_sphere.py
```diff
@@ -0,0 +1,127 @@
...
+    # Axis limits
+    xmin, xmax = -1, 1
+    ymin, ymax = -1, 1
+
+    # Particle displacement
+    amplitude = 0.05  # amplitude of the surface displacement
+    grid = sfs.util.xyz_grid([xmin, xmax], [ymin, ymax], 0, spacing=0.025)
+    ani = animation_particle_displacement(
+            omega, center, radius, amplitude, grid, frames, c='Gray')
+    ani.save('pulsating_sphere_displacement.gif', fps=10, dpi=80,
```
mgeier: You should choose a different name than in the notebook (or the other way round), otherwise one will overwrite the other's result.

doc/examples/animations_pulsating_sphere.py
```diff
@@ -0,0 +1,127 @@
...
+             writer='imagemagick')
+
+    # Particle velocity
+    amplitude = 1e-6  # amplitude of the surface displacement
+    grid = sfs.util.xyz_grid([xmin, xmax], [ymin, ymax], 0, spacing=0.04)
+    ani = animation_particle_velocity(
+            omega, center, radius, amplitude, grid, frames)
+    ani.save('pulsating_sphere_velocity.gif', fps=10, dpi=80,
```
mgeier: The same name is used in the notebook.

doc/examples/animations_pulsating_sphere.py
```diff
@@ -0,0 +1,127 @@
...
+    # Sound pressure
+    amplitude = 1e-6  # amplitude of the surface displacement
+    impedance_pw = sfs.default.rho0 * sfs.default.c
+    grid = sfs.util.xyz_grid([xmin, xmax], [ymin, ymax], 0, spacing=0.005)
+    ani = animation_sound_pressure(
+            omega, center, radius, amplitude, grid, frames, pulsate=True,
+            colorbar=True,
+            vmin=-impedance_pw * omega * amplitude,
+            vmax=impedance_pw * omega * amplitude)
+    ani.save('pulsating_sphere_pressure.gif', fps=10, dpi=80,
```
mgeier: The same name is used in the notebook.

mgeier:
Regarding https://github.com/sfstoolbox/sfs-python/pull/69#issuecomment-450841038:

> The text in the docstring is updated but the plot is not.

I found that there's an open issues for that: https://github.com/matplotlib/matplotlib/issues/12837

---

Thanks for experimenting with different grids.
That's really interesting and it looks like each one has their own weaknesses.
From your examples, I like the hexagonal layouts most.
One could try to get a "slightly less random" grid by e.g. using a hexagonal (or rectangular) layout and adding random offsets using a random range that's slightly smaller than the initial distance between particles?
Anyway, I think it's great to show multiple types of grids in the notebook to show their different properties.

The example with the random grid in the notebook seems to be broken, does that work for you?

And you should add `examples/animations-pulsating-sphere` to `doc/examples.rst`, otherwise we don't see it in the Sphinx-generated docs.

BTW, I don't see the animations anymore in the notebook, only static images. Is this intentional?
I guess not?

First I thought it's because of the `%matplotlib`, but it doesn't work without it, either.

Also, do you really want to show the static image plus the animation?

I once worked around showing the static image by calling `plt.close()`, see https://splines.readthedocs.io/en/latest/bezier.html#Preparations.

Also, as can be seen in that example, I'm quite a fan of `ani.to_jshtml()`.

It's great because it's possible to step through frame-by-frame and to change the speed by clicking the "+" and "-" buttons.

One drawback is that the notebook (including outputs) and the resulting HTML file become quite huge.
But this doesn't really matter, because those are only generated on-the-fly and are not stored in the Git repo.
It's a larger download though, if somebody looks at it online.

narahahn:
Thank you for the review.

> Anyway, I think it's great to show multiple types of grids in the notebook to show their different properties.

Yeah, I agree.

> And you should add `examples/animations-pulsating-sphere` to `doc/examples.rst`, otherwise we don't see it in the Sphinx-generated docs.

Oh, I didn't know that. Thanks. It's now added.

> BTW, I don't see the animations anymore in the notebook, only static images. Is this intentional?
> I guess not?

I wanted to include the `*.gif` files by using `` in the markdown cell,
but it didn't work the way I wanted to. It seems the notebook load the `*.gif` image
only once and store it somewhere in the memory and reuse it.
Even when new animations are generated, the displayed ones are not updated.

> Also, do you really want to show the static image plus the animation?
> 
> I once worked around showing the static image by calling `plt.close()`, see https://splines.readthedocs.io/en/latest/bezier.html#Preparations.

No, I don't. Thank you for the tip.

> Also, as can be seen in that example, I'm quite a fan of `ani.to_jshtml()`.
> 
> It's great because it's possible to step through frame-by-frame and to change the speed by clicking the "+" and "-" buttons.

That's nice. I added it in the notebook.

> You should probably pass `blit=True` and make it possible to pass in `interval` (but provide a meaningful default).

Sorry but I don't understand exactly what this is about.
The code didn't work just by adding those arguments.

> This is only used once, so what about defining it within `animation_particle_displacement()`?
> This way you could capture the variables and avoid specifying `fargs`.

You mean defining the `update_` functions within the respective `animation_` functions?
Also for `_particle_velocity` and `_pressure`, right?

doc/examples.rst
```diff
@@ -15,4 +15,5 @@ Examples
 
     examples/modal-room-acoustics
     examples/mirror-image-source-model
+	examples/animations-pulsating-sphere
```
mgeier: This uses a tab. Very bad!

narahahn: Sorry.

mgeier:
> Oh, I didn't know that. Thanks. It's now added.

You should also run Sphinx and check for errors. There are some right now.

> It seems the notebook load the *.gif image
only once and store it somewhere in the memory and reuse it.
Even when new animations are generated, the displayed ones are not updated.

Yeah, that's the browser not checking if the file was changed.
Refreshing the browser should re-load the GIF, but I agree that's a bit annoying.

If you want a nicer workflow, you could try to use something like this:

```python
from IPython.display import Image

Image('pulsating_sphere_displacement_square_grid.gif')
```

This should avoid the re-loading problem.

You should also create a normal link in some Markdown cell:

    blah [blah](pulsating_sphere_displacement_square_grid.gif) blah

This way you make sure that `nbsphinx` copies the GIF file to the target directory and visitors can easily download it.

Or you can just remove all this GIF stuff.
It's in the example script anyway.

> Sorry but I don't understand exactly what this is about.

`blit=True` is a major optimization that is used when you watch the animation in a separate plot window. It has no effect when generating a GIF and I think it hasn't an effect either when using `to_jshtml()`.

`interval` is used to specify the time (in milliseconds) between frames.

If you use this, you don't have to use `fps` when using `save()`.
And this also works when using a separate plot window.
And I hope it also works for `to_jshtml()`.

> The code didn't work just by adding those arguments.

One reason is that you removed the `return` statements from the animation function.
The other reason is that you are supposed to return a *sequence* of things, not a single thing.

> You mean defining the `update_` functions within the respective `animation_` functions?
> Also for `_particle_velocity` and `_pressure`, right?

Exactly. Just as you've done it in the meantime.

I personally would just move the function definition down a bit after all the captured variables are defined. That's not strictly necessary, but I think it makes it easier to understand where those variables come from.

narahahn:
> ```python
> from IPython.display import Image
> 
> Image('pulsating_sphere_displacement_square_grid.gif')
> ```

It works great in jupyter notebook, but unfortunately the images are not shown
in the html page generated by nbsphinx.

> One reason is that you removed the `return` statements from the animation function.
> The other reason is that you are supposed to return a _sequence_ of things, not a single thing.

Thanks, now it works. The trick was to put the returned value in a bracket `[ ]`,
which makes it iterable?

> You should also run Sphinx and check for errors. There are some right now.

I'm afraid there are still some errors.
A lot of Warnings "image file note readable" -> Is it my fault?

I guess the followings are related to my code.
```
writing output... [100%] version-history                                                                                                                                               
/home/nara/Documents/git/sfs-python/sfs/mono/source.py:docstring of sfs.mono.source.pulsating_sphere:9: WARNING: 'any' reference target not found: inside=True
/home/nara/Documents/git/sfs-python/sfs/mono/source.py:docstring of sfs.mono.source.pulsating_sphere_velocity:1: WARNING: 'any' reference target not found: inside=False
```
Can you tell what's wrong here?

mgeier:
> It works great in jupyter notebook, but unfortunately the images are not shown
in the html page generated by nbsphinx.

You mean that this is shown instead of the image:

    <IPython.core.display.Image object>

Right?

I didn't think of that! Apparently `nbconvert` doesn't support embedded GIF output, and therefore `nbsphinx` doesn't support it either. But I guess it could be enabled very easily.
But I don't really like GIF, so I guess I will not try to do that.

But if you still want to use your precious GIFs, there is yet another possible work-around:

```python
Image(url='pulsating_sphere_displacement_square_grid.gif')
```

But for this to work with `nbsphinx` you'll also need to add a Markdown-link, as I mentioned in https://github.com/sfstoolbox/sfs-python/pull/69#issuecomment-470607161.

> Thanks, now it works. The trick was to put the returned value in a bracket `[ ]`,
which makes it iterable?

Yes. It creates a Python `list` object, which is iterable.

> A lot of Warnings "image file note readable" -> Is it my fault?

Not at all!

This is a bug in Matplotlib for which I happened to have submitted a fix just a few hours ago: https://github.com/matplotlib/matplotlib/pull/13633

> Can you tell what's wrong here?

Yes, you are using single backticks where you should use double backticks.
Single backticks are configurable in reST/Sphinx. I've configured it to `any`, see http://www.sphinx-doc.org/en/master/usage/configuration.html#confval-default_role. This makes it very easy to make cross-references to other functions.

See also http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup

This is different in Markdown, where single backticks switch to code formatting.

doc/examples/animations-pulsating-sphere.ipynb
```diff
@@ -0,0 +1,341 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import sfs\n",
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "import animations_pulsating_sphere as ani\n",
+    "from IPython.display import display, Image, HTML"
```
mgeier: No need to import `display`. That's imported implicitly by recent versions of IPython.

doc/examples/animations-pulsating-sphere.ipynb
```diff
@@ -0,0 +1,341 @@
...
+    "The playback speed and direction can be changed."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "display(HTML(ani_obj.to_jshtml(default_mode='loop')))"
```
mgeier: No need to use `display()` here.

mgeier: Also, `'loop'` seems to be the default, no need to specify it here.

mgeier:
Another idea:

What about renaming the functions in `animations_pulsating_sphere.py`?

Instead of `animation_particle_displacement()` you could just use `particle_displacement()`, because the "animation" part is contained in the module name already, isn't it?

Then you could do something like:

```python
import animations_pulsating_sphere as animate
```

And later:

```python
ani = animate.particle_displacement(blah, blah, blah)
ani.save(blah, blah)
...
HTML(ani.to_jshtml())
```

I think this would look nicer, what do you think?

doc/examples/animations-pulsating-sphere.ipynb
narahahn: OK, thanks.

doc/examples/animations-pulsating-sphere.ipynb
narahahn: > I think this would look nicer, what do you think?

Yeah, good idea!

narahahn:
> I think this would look nicer, what do you think?

Yeah, good idea!

narahahn:
> But if you still want to use your precious GIFs, there is yet another possible work-around:
> 
> ```python
> Image(url='pulsating_sphere_displacement_square_grid.gif')
> ```

Now I use `to_html5_video()` to display the animations without saving them.
It also looks a bit faster than `to_jshtml()`?

Do you know how to automatically start the video in `to_jshtml()`?

> Yes, you are using single backticks where you should use double backticks.
> Single backticks are configurable in reST/Sphinx. I've configured it to `any`, see http://www.sphinx-doc.org/en/master/usage/configuration.html#confval-default_role. This makes it very easy to make cross-references to other functions.

I see, thank you!

narahahn:
> Now I use `to_html5_video()` to display the animations without saving them.
> It also looks a bit faster than `to_jshtml()`?

@mgeier Or wo

> 
> Do you know how to automatically start the video in `to_jshtml()`?
> 
> > Yes, you are using single backticks where you should use double backticks.
> > Single backticks are configurable in reST/Sphinx. I've configured it to `any`, see http://www.sphinx-doc.org/en/master/usage/configuration.html#confval-default_role. This makes it very easy to make cross-references to other functions.
> 
> I see, thank you!

narahahn:
@mgeier Would you prefer just using `to_jshtml()` for all animations?

mgeier:
> Would you prefer just using `to_jshtml()` for all animations?

You should decide, I like both.

I just don't like GIFs, but `to_html5_video()` is fine for me.

Whatever you choose, it would be great to show the other option in one example (as you are currently doing).

BTW, converting the animation to a video doesn't currently work on Travis-CI. I think you can fix this by adding `ffmpeg` to the requirements in `.travis.yml`:

```yaml
addons:
  apt:
    packages:
      - pandoc
      - ffmpeg
```

narahahn:
> I just don't like GIFs, but `to_html5_video()` is fine for me.

I prefer `to_html5_video()` just because the video starts automatically in the notebook.
Unfortunately, this doesn't seems to be the case in sphinx.
BTW, what's wrong with GIF?

> Whatever you choose, it would be great to show the other option in one example (as you are currently doing).

I agree.

> BTW, converting the animation to a video doesn't currently work on Travis-CI. I think you can fix this by adding `ffmpeg` to the requirements in `.travis.yml`:

Thanks.

mgeier:
> Now I use `to_html5_video()` to display the animations without saving them.
> It also looks a bit faster than `to_jshtml()`?

Yes, it's definitely much faster to compute. And when saving it to the notebook the resulting notebook is much smaller.

> Do you know how to automatically start the video in `to_jshtml()`?

No. I think it's not possible. I would also like to have that.
This could be a feature request for matplotlib?

> I prefer `to_html5_video()` just because the video starts automatically in the notebook.
> Unfortunately, this doesn't seems to be the case in sphinx.

This works for me, but some of the videos just randomly stopped at some point (probably when switching between browser tabs?).
Probably the browser has problems playing that many videos at the same time?
After refreshing the browser page, the videos were playing again ...

> BTW, what's wrong with GIF?

It's an outdated technology and it was for a long time encumbered by patents (which it isn't anymore, but it still has this slightly ugly taste to it).

mgeier:
FYI: some tests on Travis-CI don't pass, that's the fix: #118

doc/examples/animations-pulsating-sphere.ipynb
```diff
@@ -0,0 +1,326 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import sfs\n",
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "import animations_pulsating_sphere as animation\n",
```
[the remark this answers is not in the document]
narahahn: Is it possible to make a link that works in sphinx as well as in jupyter notebook?

doc/examples/animations-pulsating-sphere.ipynb
```diff
@@ -0,0 +1,326 @@
...
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "amplitude = 1e-6\n",
+    "impedance_pw = sfs.default.rho0 * sfs.default.c\n",
+    "max_pressure = omega * impedance_pw * amplitude\n",
+    "\n",
+    "grid = sfs.util.xyz_grid([xmin, xmax], [ymin, ymax], 0, spacing=0.005)\n",
+    "ani = animation.sound_pressure(\n",
+    "        omega, center, radius, amplitude, grid, frames, pulsate=False,\n",
```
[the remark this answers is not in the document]
narahahn: Yes, you're right.

doc/examples/animations_pulsating_sphere.py
```diff
@@ -0,0 +1,117 @@
...
+               omega, center, radius, amplitude, grid)
+    displacement = sfs.util.displacement(velocity, omega)
+    phasor = np.exp(1j * 2 * np.pi / frames)
+
+    def update_frame_displacement(i):
+        position = (grid + displacement * phasor**i).apply(np.real)
+        position = np.column_stack([position[0].flatten(),
+                                    position[1].flatten()])
+        scat.set_offsets(position)
+        return [scat]
```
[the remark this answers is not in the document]
narahahn: Done.

narahahn:
> FYI: some tests on Travis-CI don't pass, that's the fix: #118

Should I apply it in this PR?

doc/examples/animations-pulsating-sphere.ipynb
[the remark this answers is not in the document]
mgeier: Yes it is. But it's not possible to use inline formatting within a link in Sphinx (or more exactly, in reStructuredText, which is currently an intermediate format used in `nbsphinx`), see http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#gotchas.

IOW, instead of

    [`animations_pulsating_sphere.py`](animations_pulsating_sphere.py)

... you'll have to use

    [animations_pulsating_sphere.py](animations_pulsating_sphere.py)

... which isn't that bad IMHO.

doc/examples/animations-pulsating-sphere.ipynb
```diff
@@ -0,0 +1,363 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this example, the sound field of a pulsating sphere is visualized.\n",
+    "Different acoustic variables, such as sound pressure,\n",
+    "particle velocity, and particle displacement, are simulated.\n",
+    "The first two quantities are computed with\n",
+    "\n",
+    "- `sfs.mono.source.pulsating_sphere()` and\n",
```
mgeier: If you want to be fancy, you can use links like this:

    [sfs.mono.source.pulsating_sphere()](../sfs.mono.source.rst#sfs.mono.source.pulsating_sphere)

This isn't really meaningful in JupyterLab, but it links to the documentation in Sphinx's HTML output.

mgeier:
> > FYI: some tests on Travis-CI don't pass, that's the fix: #118
>
> Should I apply it in this PR?

No need. Just ignore the CI results for Python 3.5 and 3.6.

mgeier:
Sorry, I forgot something in my previous review:

```
$ python3 -m pydocstyle doc/examples/animations_pulsating_sphere.py 
doc/examples/animations_pulsating_sphere.py:8 in public function `particle_displacement`:
        D202: No blank lines allowed after function docstring (found 1)
doc/examples/animations_pulsating_sphere.py:33 in public function `particle_velocity`:
        D202: No blank lines allowed after function docstring (found 1)
doc/examples/animations_pulsating_sphere.py:55 in public function `sound_pressure`:
        D202: No blank lines allowed after function docstring (found 1)
```

narahahn:
> Sorry, I forgot something in my previous review:

Sorry, I should have checked it.

> If you want to be fancy, you can use links like this:

Much better, thank you.

It's a bit unfortunate that the links don't look like those in the API document.

> It's an outdated technology and it was for a long time encumbered by patents (which it isn't anymore, but it still has this slightly ugly taste to it).

Would it be better to save the animations of the python script example
in a different format?

mgeier:
> It's a bit unfortunate that the links don't look like those in the API document.

What do you mean by that?

> Would it be better to save the animations of the python script example in a different format?

No, that's OK.

We can add other formats later, if we feel like it.

narahahn:
> What do you mean by that?

I was hoping to make a link to the functions with programming font inside a box.
The same appearance as the links in API doc.

> We can add other formats later, if we feel like it.

Alright.

> Do you want to keep all the individual commits in the history?
> Or do you want to squash some or all of them?

No, I will squash and rebase when it's ready to merge.

mgeier:
> I was hoping to make a link to the functions with programming font inside a box.

OK, I see. That's currently not possible.
It should become possible to use nested markup when https://github.com/spatialaudio/nbsphinx/issues/36 is implemented (which will not be soon), but even then, it might not look exactly like the links in the rest of the Sphinx pages.

> No, I will squash and rebase when it's ready to merge.

I think it's ready!

narahahn:
> I think it's ready!

Here we go.

Thanks a lot for your review.

Issue #59, opened 2018-04-30
alpha not in [0, 2\pi)


chris-hld:
https://github.com/sfstoolbox/sfs-python/blob/ac86493b1a07775731c46d4d1701660a6e83258d/sfs/util.py#L125

alpha is not in [0, 2\pi) but in [-pi, pi].
This can turn out as a bug when using it.

We could add
`alpha[alpha < 0] += 2*np.pi`
or something even smarter.

chris-hld:
This is be better and should be faster:
`alpha = (alpha + 2*np.pi) % (2*np.pi)`

mgeier:
Why should it be the one range and not the other?

Is there any math textbook or relevant software that uses your suggested behavior for azimuth?

> This can turn out as a bug when using it.

But that's a bug in the other code then, isn't it?
Any function taking an angle in radians should simply do *The Right Thing* when given an "out-of-range" value.

A function that only takes values from -pi to pi is equally broken.

hagenw:
Interesting discussion. I agree that all functions should work with out of range values as input.

For return values I would expect that they are within the range used throughout the toolbox. At the moment this would be `[0, 2pi)` as this is used in the theory at the moment, see http://sfstoolbox.org/en/2.9.0/#coordinate-system.
The same convention is used by [Ahrens 2012](https://doi.org/10.1007/978-3-642-25743-8) and [Schultz 2016](http://rosdok.uni-rostock.de/resolve/urn/urn:nbn:de:gbv:28-diss2016-0078-1).

So maybe, we should not change the documentation, but the return value.

mgeier:
@hagenw Thanks for the references, that's very helpful!

Ahrens 2012 in turn references this: Weisstein, E. W. (2002). CRC concise encyclopedia of mathematics. London: Chapman and Hall/CRC. I don't have access to this book, so I couldn't check it out.

I agree that the range `[0, 2pi)` makes sense in a theoretical math context. I'm not sure, however, if it also makes sense in a numeric implementation.
The `numpy.arctan2()` has a nice implementation of some "special" values (as shown in the documentation linked by @chohner), which we would most likely destroy if we add our own number crunching.

So I would still be interested how other software does this!

fs446:
Here is a reference from Wolfram, which due to collaboration of the parties might be consistent with the Weisstein book (however I did not check):
http://functions.wolfram.com/PDF/ArcTan2.pdf

[...]

PR #73, opened 2019-01-07
Fix cart2sph to reflect correct range and update tests.

chris-hld:
This is the PR for https://github.com/sfstoolbox/sfs-python/issues/59.
Testing against `np.arctan2` is not meaningful, since this is causing the range issue.

### DIFF
--- a/sfs/util.py
+++ b/sfs/util.py
@@ -123,6 +123,7 @@ def cart2sph(x, y, z):
     """
     r = np.sqrt(x**2 + y**2 + z**2)
     alpha = np.arctan2(y, x)
+    alpha = alpha % (2 * np.pi)  # [0, 2pi)
     beta = np.arccos(z / r)
     return alpha, beta, r
 
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -4,14 +4,18 @@ import pytest
 import sfs
 
 cart_sph_data = [
-    ((1, 1, 1), (np.pi/4, np.arccos(1/np.sqrt(3)), np.sqrt(3))),
-    ((-1, 1, 1), (np.arctan2(1, -1), np.arccos(1/np.sqrt(3)), np.sqrt(3))),
-    ((1, -1, 1), (-np.pi/4, np.arccos(1/np.sqrt(3)), np.sqrt(3))),
-    ((-1, -1, 1), (np.arctan2(-1, -1), np.arccos(1/np.sqrt(3)), np.sqrt(3))),
-    ((1, 1, -1), (np.pi/4, np.arccos(-1/np.sqrt(3)), np.sqrt(3))),
-    ((-1, 1, -1), (np.arctan2(1, -1), np.arccos(-1/np.sqrt(3)), np.sqrt(3))),
-    ((1, -1, -1), (-np.pi/4, np.arccos(-1/np.sqrt(3)), np.sqrt(3))),
-    ((-1, -1, -1), (np.arctan2(-1, -1), np.arccos(-1/np.sqrt(3)), np.sqrt(3))),
+    ((1, 1, 1), (np.pi / 4, np.arccos(1 / np.sqrt(3)), np.sqrt(3))),
+    ((-1, 1, 1), (3 * np.pi / 4, np.arccos(1 / np.sqrt(3)), np.sqrt(3))),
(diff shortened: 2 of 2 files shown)


mgeier:
Thanks for this PR and sorry for the late response!

As mentioned in https://github.com/sfstoolbox/sfs-python/issues/59#issuecomment-388902384, this destroys a few of the "special" values mentioned in https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.arctan2.html.

For example, those are (first result is the previous behavior, second result is this PR):

```python
import numpy as np
import sfs

sfs.util.cart2sph(0, -0., 1)[0]
# -0.0
# 0.0
sfs.util.cart2sph(np.inf, -0., 0)[0]
# -0.0
# 0.0
sfs.util.cart2sph(-0., -0., 1)[0]
# -3.141592653589793
# 3.141592653589793
sfs.util.cart2sph(-np.inf, -0., 0)[0]
# -3.141592653589793
# 3.141592653589793
```

I think from a geometrical point of view, the old values make a lot of sense, don't they?

chris-hld:
Ok, now I see your point more clearly. 

Here is my opinion on this:
We state that the azimuth is in `[0, 2pi)` . This is getting a bit philosophically speaking, but this interval does not include `-0`. So, if any of this, `(0, -0., 1)` may return `2pi`, which is not included in the domain, though.
For the last two examples I'd say generally the second value is as correct as the first. Again, our domain is `[0, 2pi)`, so the limit is `pi`, not `-pi`, the latter is simply not approachable. The explanation is that we are approaching pi from the positive direction 2pi. 
This is just different than atan2, which approaches this value continuously from `-0`, because it has a different domain. To my knowledge this is actually not the same, neither is this a special value.
So in short:
atan2 does `-0 -> -pi/2 -> -pi`.
We should do `2pi -> 3/2 pi -> pi`.
atan2 is discontinuous in `pi/-pi`, we are in `0/2pi`.
Both values are correct, but not the same.

Honestly, these are special values that we never actually use or treat separately, are we?
Are we catching these special values somewhere?
Otherwise it's pretty confusing if we say our values are within a certain range in the docs but we actually don't provide them. However, I agree that it shouldn't make a difference if any following implementation is correct.

mgeier:
> We state that the azimuth is in `[0, 2pi)`

Yes we do, but we probably shouldn't (as discussed in #59).

As mentioned in #59, any function that doesn't work with the full input range of real numbers is broken.
So for all practical use cases we have considered so far, it doesn't matter one bit what range we return.

But of course we have to choose some range.
Above I've given some motivation why the range `[-pi, pi]` makes sense.
Incidentally, that also leads to the simplest implementation.
I don't see a reason why we should jump through hoops in our implementation for zero gain.

In a theoretical math context another range may or may not make more sense, but I don't see a problem in using two different ranges for theoretical and practical work.

Of course, it should be documented correctly.

> For the last two examples I'd say generally the second value is as correct as the first.

That may be the case for some definitions of "correct", but the second value definitely throws away information compared to the first value.

> Again, our domain is `[0, 2pi)`

Again, it doesn't have to be.

>  To my knowledge this is actually not the same, neither is this a special value.

I don't understand this sentence, can you please clarify?

> Honestly, these are special values that we never actually use or treat separately, are we?

Probably not. But I don't see a reason why we should throw away this information.

> Are we catching these special values somewhere?

I guess not. At least not yet.

But since we are implementing a library here, we don't have all possible use cases available to us.

In this case I think we should try to do The Right Thing.

> Otherwise it's pretty confusing if we say our values are within a certain range in the docs but we actually don't provide them.

I agree. That's why we should change the documentation, see https://github.com/sfstoolbox/sfs-python/issues/59#issuecomment-386634275.

Unless somebody provides a real reason to change the behavior.

chris-hld:
> But since we are implementing a library here, we don't have all possible use cases available to us.
>In this case I think we should try to do The Right Thing.

I agree. 
So then, let's only update the docs. I personally think it is still better to test against pre-calculated values, and not the return of `np.arctan2()` (as the current test does). What do you think?

chris-hld:
Closing and moving back to issue #59.

mgeier:
> I personally think it is still better to test against pre-calculated values, and not the return of `np.arctan2()` (as the current test does). What do you think?

I agree.

Would you mind extracting your changes to the tests from this PR and making a new PR only for the tests?

PR #74, opened 2019-01-09
Fix db for power quantities

narahahn:
`db(x)` is always returning `10` for `power=True`.


narahahn:
Oh, nice! I didn't know that.

PR #76, opened 2019-01-20
 Drop support for legacy Python 2

### LINKED ISSUE #71: Drop Python 2 support?
Anyone against that?

hugovk:
Fixes #71.


mgeier:
@hugovk Thanks a lot for this PR!

Is it really necessary to support 3.6 and 3.7 specifically?

```python
    python_requires='>=3.6',
```

What about using this instead?

```python
    python_requires='>=3',
```

And what about dropping those two lines:

```python
        "Programming Language :: Python :: 3.6",
        "Programming Language :: Python :: 3.7",
```

Are those really helpful?
I'm quite sure we'll forget to update them for Python 3.8 etc.

hugovk:
You're welcome!

I wasn't sure which versions are officially supported, as only 3.6 is tested.

python_requires limits which versions pip will install for, so it's better for this range to be wider if you're not sure.

Which versions of 3.x are supported? If anything, I'd recommend 3.4+, as all the others are EOL. (And 3.4 is EOL in couple of months.)

Those two Trove classifiers are useful, it shows on PyPI which versions are known to work. But I will remove them if you prefer.

mgeier:
> Which versions of 3.x are supported?

TBH, I don't know exactly. I was assuming >= 3.4, and probably 3.3. Anything less than 3.3 was quite broken anyway ...

Would you mind adding 3.5 and 3.7 to the Travis-CI config?
Then we can see if it passes ...

If you like we can try 3.4, too, but I'm fine with dropping support for 3.4. Support for 3.5 would be good, though.

> Those two Trove classifiers are useful

OK, then we'll keep them!
Can you please add 3.5 then?

hugovk:
Updated and all green!

chris-hld:
I strongly vote for Python >= 3.5, mostly because of the `@` operator, which we could then start using.
https://docs.python.org/3/whatsnew/3.5.html#whatsnew-pep-465

PR #79, opened 2019-02-10
TST: Make script to run all test scripts; use it on Travis-CI

mgeier:


spors:
The script is quite specific and can be removed from the examples.

PR #80, opened 2019-02-10
DOC: Make LaTeX math definitions LaTeX compatible

mgeier:
Our math definitions cause problems when creating LaTeX output because they are not actually LaTeX compatible, they just happen to work with MathJax.

There are two reasons why this doesn't work in LaTeX:

* the definitions are included multiple times, leading to `\newcommand` complaining about already defined macros
* While in MathJax we need to be in math mode for anything to work, in LaTeX the `\newcommand` doesn't work in math mode.

My work-around is admittedly a bit non-obvious, does anyone have a better idea how to fix this?

### DIFF
--- a/doc/math-definitions.rst
+++ b/doc/math-definitions.rst
@@ -1,10 +1,22 @@
+.. raw:: latex
+
+    $
+
 .. rst-class:: hidden
 .. math::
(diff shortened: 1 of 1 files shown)


mgeier:
I've merged #90 instead.

PR #82, opened 2019-02-11
Fix LaTeX math definitions

mgeier:
An alternative to #80 and #81.

I had to change `\newcommand{\i}` to `\renewcommand{\i}`, it seemed to be defined already ...


mgeier:
I've merged #90 instead.

Issue #32, opened 2016-12-20
Rename "defs"?


mgeier:
The submodule `sfs.defs` is supposed to mean "defaults".
This is quite accurate, since it contains default values.

However, if we would rename it to `default`, it would make more sense when reading it, e.g. `sfs.default.c` reads like "default c", not like "defaults c".

Apart from renaming, we could also turn it into an object of a custom class, like e.g. http://python-sounddevice.readthedocs.io/#sounddevice.default
This way, we could avoid accidentally assigning unsupported attributes.

PR #84, opened 2019-02-11
Use sfs.default object instead of sfs.defs submodule

mgeier:
See #32.


hagenw:
I like your changes :+1: 

Would be great if you could integrate them in the documentation as well. I think it would be nice if we tell a user how she/he could change the default speed of sound.

mgeier:
What about 8564d71a70aa7f76e835bd3f9269e2a1fad2ab1c?

I was hesitating to add this to the docs, because it's a bit messy.

Does anyone have an idea how to do this in a cleaner way?

mgeier:
#45 and #69 should be merged before this PR because they use `sfs.defs` a few times.

mgeier:
I've moved the `autosummary`directive in a4c7c8c2d35f130bb4768e63af20f99df509d166.

This way, `sfs.default` is documented directly on the main API page and not on a separate page.

I think that's better.

Any other opinions?

PR #88, opened 2019-02-12
DOC: math notations for mono sources

hagenw:
This replaces the ASCII math notations for the monochromatic source definitions.


mgeier:
Thanks @hagenw!

PR #90, opened 2019-02-18
DOC: Make LaTeX math definitions LaTeX compatible

mgeier:
Yet another alternative for #80, #81 and #82.

PR #92, opened 2019-02-21
Doc tweaks

mgeier:


hagenw:
Looks good to me. Tested HTML locally, PDF still fails as #90 is not merged.

When I experimented with SVG figures in HTML output some years ago it resulted in heavy CPU usage and long load time if you had several sound fields on one page, but this seems to be no longer a problem.

Issue #75, opened 2019-01-09
Sound field plot using pyplot.imshow()


narahahn:
When `pyplot.imshow()` is used without `extent` option, the pixels are shown on integer coordinates.
![imshow](https://user-images.githubusercontent.com/8125968/50912195-d752ed80-1431-11e9-9138-c2e8c7caae77.png)

The position of the pixels can be adjusted by using `extent` as we are doing in `sfs.plot.soundfield`. Below I choose `extent=[-1, 2, -1, 1]`
![imshow-extent](https://user-images.githubusercontent.com/8125968/50912353-42042900-1432-11e9-9416-752968e3252a.png)

All pixels are fit into the defined range. As a result, the plot is a bit smaller than it should be. You can see that the pixels are off the grid points. For instance, the lower left pixel is supposed to be centered at [-1, 1] but is now inside the defined range.
In most cases, this won't be a big problem for high-resolution plots. But if we want to draw lines or add patches on top of the plot, it might be noticeable. Here is an example.
![pulsating-sphere-pressure](https://user-images.githubusercontent.com/8125968/50913102-e9359000-1433-11e9-8ac9-c49933890cca.png)

The origin of the sound field and the center of the circle do not match exactly.

One way to circumvent this issue is changing the `extent` values,
`extent=[xmin, xmax, ymin, ymax]`
to
`extent=[xmin - dx / 2, xmax + dx / 2, ymin - dy / 2, ymax + dy / 2]`
where `dx` and `dy` are the spacing of the grid in the respective coordinates.
![imshow-corrected-extent](https://user-images.githubusercontent.com/8125968/50913686-277f7f00-1435-11e9-8bfd-e50002107c78.png)

Alternatively, we can use [`pyplot.pcolormesh`](https://matplotlib.org/api/_as_gen/matplotlib.pyplot.pcolormesh.html) instead.
It might be preferred since it takes the x- and y-axes (let say `x` and `y`) as input arguments, so there will be no scaling issue.
A downside for our use case is that the pixels are not shown at (`x[i]`, `y[i]`) but inside a rectangle defined by (`x[i]`, `x[i+1]`, `y[i]`, `y[i+1]`). So the `x` and `y` should have one more element than the dimensions of the data. Otherwise, the image will be cropped like this.
![pcolormesh](https://user-images.githubusercontent.com/8125968/50914451-d96b7b00-1436-11e9-9230-a8e51ced653f.png)

[...]

PR #96, opened 2019-02-28
Adjust axis limits in sound field plots

narahahn:
Before
soundfield_before

After
soundfield_after

These plots are generated by

        sfs-python/doc/examples/sound_field_synthesis.py
      
with 

```python
grid = sfs.util.xyz_grid([-2, 2], [-2, 2], 0, spacing=0.5)
```

### DIFF
--- a/sfs/plot.py
+++ b/sfs/plot.py
@@ -283,6 +283,9 @@ def soundfield(p, grid, xnorm=None, cmap='coolwarm_clip', vmin=-2.0, vmax=2.0,
     elif plotting_plane == 'yz':
         x, y = grid[[1, 2]]
 
+    dx = np.asscalar(np.diff(x[0, :2])) / 2
+    dy = np.asscalar(np.diff(y[:2, 0])) / 2
+
     if ax is None:
         ax = plt.gca()
 
@@ -291,7 +294,7 @@ def soundfield(p, grid, xnorm=None, cmap='coolwarm_clip', vmin=-2.0, vmax=2.0,
         p = np.clip(p, -1e15, 1e15)  # clip to float64 range
 
     im = ax.imshow(np.real(p), cmap=cmap, origin='lower',
-                   extent=[x.min(), x.max(), y.min(), y.max()],
+                   extent=[x.min()-dx, x.max()+dx, y.min()-dy, y.max()+dy],
                    vmax=vmax, vmin=vmin, **kwargs)
     if xlabel is None:
         xlabel = plotting_plane[0] + ' / m'
--- a/sfs/plot.py
+++ b/sfs/plot.py
@@ -283,8 +283,8 @@ def soundfield(p, grid, xnorm=None, cmap='coolwarm_clip', vmin=-2.0, vmax=2.0,
     elif plotting_plane == 'yz':
         x, y = grid[[1, 2]]
 
-    dx = np.asscalar(np.diff(x[0, :2])) / 2
-    dy = np.asscalar(np.diff(y[:2, 0])) / 2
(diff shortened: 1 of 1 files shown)


narahahn:
This PR fixes the issue #75.

sfs/plot.py
```diff
@@ -281,6 +281,9 @@ def soundfield(p, grid, xnorm=None, cmap='coolwarm_clip', vmin=-2.0, vmax=2.0,
     elif plotting_plane == 'yz':
         x, y = grid[[1, 2]]
 
+    dx = np.asscalar(np.diff(x[0, :2])) / 2
+    dy = np.asscalar(np.diff(y[:2, 0])) / 2
```
mgeier: This looks like it makes more assumptions than it should.

Does this work when using a vertical grid slice?

For example using `sfs.util.xyz_grid()` with `x=0` or `y=0` instead of `z=0`?

narahahn: Yes, it was tested for `xz` plane and `yz` plane.

soundfield_xz
soundfield_yz

In the following part of the code,
_nonscalar_ axes are defined as `x` and `y`,
which are always column and row vectors, respectively.

https://github.com/sfstoolbox/sfs-python/blob/c10fa334eb832f707078a2bd626e9746c1607b84/sfs/plot.py#L277-L282

This will work only for equidistant girds, though.
If we want to plot a sound field on a nonuniform gird,
we should figure out something else.
BTW, I don't know if `imshow` or `pcolormesh` supports non-uniform grids,

mgeier: Thanks for testing this!

> which are always column and row vectors, respectively.

Yes, but only when using `sfs.util.xyz_grid()`.

However, it doesn't work when using `numpy.meshgrid()`, for example:

```python
grid = np.meshgrid(
    [-2, -1.5, -1, -0.5, 0, 0.5, 1, 1.5, 2],
    [-2, -1.5, -1, -0.5, 0, 0.5, 1, 1.5, 2],
    [0],
    #sparse=True,
)
```

... which without your change works (with and without `sparse`) and should result in the same plot.

I think a more robust (and shorter, and easier-to-understand) solution would be something like this:

```python
dx = 0.5 * x.ptp() / p.shape[1]
dy = 0.5 * y.ptp() / p.shape[0]
```

This doesn't make any assumptions about the shape of `x` and `y` (except that they must not be empty).

> If we want to plot a sound field on a nonuniform gird,
> we should figure out something else.

I don't think that `imshow()` does the right thing on nonuniform grids. It has to assume uniformity.

I think we should make calculations work with non-uniform grids, but I think it's fine to assume uniform grids for this kind of plotting functions.
We could at some point add plotting functions for point clouds similar to http://matlab.sfstoolbox.org/en/2.4.3/custom-grid/, but those should be separate functions.

BTW, calculating those point clouds should already work (but I haven't thoroughly tested this).

mgeier: We should probably add the uniformity assumption to the docs?

narahahn: > ```python
> dx = 0.5 * x.ptp() / p.shape[1]
> dy = 0.5 * y.ptp() / p.shape[0]
> ```

Yeah, that's much better. Thanks.
I will take this over.

> BTW, calculating those point clouds should already work (but I haven't thoroughly tested this).

The data should then be plotted using `scatter` I guess?

narahahn: > We should probably add the uniformity assumption to the docs?

Would that be necessary?
`imshow()` isn't suited for nonuniform grid anyway.

mgeier: > Would that be necessary?

I don't know.
Probably not. But it wouldn't hurt either ...

PR #97, opened 2019-02-28
docu: examples for sfs.array

fs446:
added examples for array functions
need further changes when loudspeaker_2d,_3d handling is optimized
please double check the bugfix in planar() where N1 and N2 were exchanged


sfs/array.py
```diff
@@ -423,16 +460,38 @@ def cube(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
 
 
 def sphere_load(fname, radius, center=[0, 0, 0]):
-    """Spherical secondary source distribution loaded from datafile.
+    """Spherical secondary source distribution loaded from the datafile fname.
```
mgeier: I don't really like this.

If you refer to a parameter, you should use some markup, I would suggest `*fname*`.

But actually I don't really like the name `fname`. It's hideous.

I know that's taken from NumPy (https://docs.scipy.org/doc/numpy/reference/generated/numpy.loadtxt.html), but it's bad nonetheless.

I would prefer changing the parameter name to `file`.

Then the tagline could be:

    Spherical secondary source distribution loaded from *file*.

But that's still bad.
This would be better:

    Load spherical secondary source distribution from *file*.

Or drop the markup in this case. Both is possible.

sfs/array.py
```diff
@@ -423,16 +460,38 @@ def cube(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
...
 
-    ASCII Format (see MATLAB SFS Toolbox) with 4 numbers (3 position, 1
-    weight) per secondary source located on the unit circle.
+    ASCII Format (see MATLAB SFS Toolbox) with 4 numbers (3 for the cartesian 
+    position vector, 1 for the integration weight) per secondary source located 
+    on the unit circle which is resized by the set radius and shifted to the
```
mgeier: `*radius*` and `*center*`.

I think "given" would be better than "set", but I'm no native speaker ...

sfs/array.py
```diff
@@ -423,16 +460,38 @@ def cube(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
...
+        :context: close-figs
+
+        #content of 'test_sphere.txt' is the trivial t-design of
+        #http://neilsloane.com/sphdesigns/dim3/des.3.6.3.txt:
+        #1 0 0 1
+        #-1 0 0 1
+        #0 1 0 1
+        #0 -1 0 1
+        #0 0 1 1
+        #0 0 -1 1
```
mgeier: What about moving those comments out of the plotting code and into the description before (or after) the code?

Then you could use an actual URL and could show the values in nice formatting.

sfs/array.py
```diff
@@ -443,16 +502,33 @@ def sphere_load(fname, radius, center=[0, 0, 0]):
 
 
 def load(fname, center=[0, 0, 0], orientation=[1, 0, 0]):
-    """Load secondary source positions from datafile.
+    """Load arrangement of secondary sources from the datafile fname.
```
mgeier: Same as above.

sfs/array.py
```diff
@@ -443,16 +502,33 @@ def sphere_load(fname, radius, center=[0, 0, 0]):
...
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        #content of 'test_array.csv' is:
+        #1,0,0,-1,0,0,1
+        #0,1,0,0,-1,0,1
+        #-1,0,0,1,0,0,1
+        #0,-1,0,0,1,0,1
```
mgeier: Same as above.

sfs/array.py
```diff
@@ -484,6 +560,15 @@ def weights_midpoint(positions, closed):
     (N,) numpy.ndarray
         Weights of secondary sources.
 
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        x0, n0, a0 = sfs.array.circular(2**8, 1)
+        a = sfs.array.weights_midpoint(x0, closed='True')
```
mgeier: `'True'` happens to be `True`, but you should still use `closed=True`.

sfs/array.py
```diff
@@ -484,6 +560,15 @@ def weights_midpoint(positions, closed):
...
+        print(max(abs(a0-a)))  # check error against analytic solution, i.e. 6.160347305397829e-07
```
mgeier: The output of `print()` is not shown when using the plot directive.

You could use doctests, like e.g. in http://python.sfstoolbox.org/en/0.4.0/utilities.html#sfs.util.as_delayed_signal

Also, the line is too long.

fs446:
> IMHO the file names of the two new files are not great.
> 
> Could you probably choose something a bit more descriptive?

changed to 
example_array_4LS_2D.csv
example_array_6LS_3D.txt

sfs/array.py
fs446: changed handling to
Load spherical secondary source distribution from file.

sfs/array.py
fs446: changed to "given"

sfs/array.py
fs446: good idea, was not thinking about that option when implementing this in the first instance. Changed to the suggested version.

sfs/array.py
fs446: changed

sfs/array.py
fs446: as above

sfs/array.py
fs446: old habits, changed to True

sfs/array.py
fs446: nice tool, thanks! changed to this handling

fs446:
please double check changing rostock array data:
https://github.com/sfstoolbox/sfs-python/commit/2599600b41aec33fec8f5c80711534d18040991f

sfs/array.py
```diff
@@ -15,7 +15,7 @@
 
 
 class ArrayData(namedtuple('ArrayData', 'x n a')):
-    """Named tuple returned by array functions.
+    """Get named tuple returned by array functions.
```
mgeier: This is not a function, therefore the docstring was more appropriate before.

sfs/array.py
```diff
@@ -43,7 +43,7 @@ def take(self, indices):
 
 
 def linear(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
-    """Linear secondary source distribution.
+    """Get linear, equidistantly sampled secondary source distribution.
```
mgeier: This is better than before, but probably we should use "Return" instead of "Get"?

See https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

sfs/array.py
```diff
@@ -135,6 +139,8 @@ def linear_random(N, min_spacing, max_spacing, center=[0, 0, 0],
         x0, n0, a0 = sfs.array.linear_random(12, 0.15, 0.4, orientation=[0, -1, 0])
         sfs.plot.loudspeaker_2d(x0, n0, a0)
         plt.axis('equal')
+        plt.xlabel('x / m')
+        plt.ylabel('y / m')
```
mgeier: sfs/array.py:143:28: W291 trailing whitespace

Please enable https://editorconfig.org/ in your editor.

sfs/array.py
```diff
@@ -229,7 +239,7 @@ def rectangular(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
 
 
 def rounded_edge(Nxy, Nr, dx, center=[0, 0, 0], orientation=[1, 0, 0]):
-    """Array along the xy-axis with rounded edge at the origin.
+    """Get secondary source distribution along the xy-axis with rounded edge at the origin.
```
mgeier: sfs/array.py:242:80: E501 line too long (91 > 79 characters)

sfs/array.py
```diff
@@ -301,7 +313,7 @@ def rounded_edge(Nxy, Nr, dx, center=[0, 0, 0], orientation=[1, 0, 0]):
 
 
 def edge(Nxy, dx, center=[0, 0, 0], orientation=[1, 0, 0]):
-    """Array along the xy-axis with edge at the origin.
+    """Get secondary source distribution along the xy-axis with sharp edge at the origin.
```
mgeier: sfs/array.py:316:80: E501 line too long (89 > 79 characters)

sfs/array.py
```diff
@@ -422,39 +459,97 @@ def cube(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
...
 
 
-def sphere_load(fname, radius, center=[0, 0, 0]):
-    """Spherical secondary source distribution loaded from datafile.
+def sphere_load(file, radius, center=[0, 0, 0]):
+    """Load spherical secondary source distribution from file.
 
-    ASCII Format (see MATLAB SFS Toolbox) with 4 numbers (3 position, 1
-    weight) per secondary source located on the unit circle.
+    ASCII Format (see MATLAB SFS Toolbox) with 4 numbers (3 for the cartesian
```
mgeier: sfs/array.py:465:78: W291 trailing whitespace

sfs/array.py
```diff
@@ -422,39 +459,97 @@ def cube(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
...
 
-def sphere_load(fname, radius, center=[0, 0, 0]):
-    """Spherical secondary source distribution loaded from datafile.
+def sphere_load(file, radius, center=[0, 0, 0]):
+    """Load spherical secondary source distribution from file.
 
-    ASCII Format (see MATLAB SFS Toolbox) with 4 numbers (3 position, 1
-    weight) per secondary source located on the unit circle.
+    ASCII Format (see MATLAB SFS Toolbox) with 4 numbers (3 for the cartesian 
+    position vector, 1 for the integration weight) per secondary source located
```
mgeier: sfs/array.py:466:80: W291 trailing whitespace

sfs/array.py
```diff
@@ -422,39 +459,97 @@ def cube(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
...
+    given center.
 
     Returns
     -------
     `ArrayData`
         Positions, orientations and weights of secondary sources.
 
+    Examples
+    --------
+    content of `example_array_6LS_3D.txt`
```
mgeier: sfs/array.py:477:42: W291 trailing whitespace

sfs/array.py
```diff
@@ -422,39 +459,97 @@ def cube(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
...
+        0 0 1 1
+        0 0 -1 1
+
+    corresponds to the `3-dimensional 6-point spherical 3-design
+    <http://neilsloane.com/sphdesigns/dim3/des.3.6.3.txt>`_.
+
+    .. plot::
+        :context: close-figs
+
+        x0, n0, a0 = sfs.array.sphere_load('../data/arrays/example_array_6LS_3D.txt', radius=2, center=[0, 0, 0])
```
mgeier: sfs/array.py:492:80: E501 line too long (113 > 79 characters)

In this case it's OK to exceed the line length in the source file because there are 8 spaces indentation which don't show up in the generated docs, but you should not exceed 87 characters in this case.

sfs/array.py
```diff
@@ -422,39 +459,97 @@ def cube(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
...
+        sfs.plot.loudspeaker_2d(x0, n0, a0, size=0.25)
+        plt.axis('equal')
+        plt.xlabel('x / m')
+        plt.ylabel('y / m')
+        plt.title('view onto xy-plane')
```
mgeier: sfs/array.py:497:40: W291 trailing whitespace

sfs/array.py
```diff
@@ -422,39 +459,97 @@ def cube(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
...
 
-def load(fname, center=[0, 0, 0], orientation=[1, 0, 0]):
-    """Load secondary source positions from datafile.
+def load(file, center=[0, 0, 0], orientation=[1, 0, 0]):
+    """Load secondary source distribution from file.
 
-    Comma Seperated Values (CSV) format with 7 values
-    (3 positions, 3 normal vectors, 1 weight) per secondary source.
+    Comma Separated Values (CSV) format with 7 values
+    (3 for the cartesian position vector, 3 for the cartesian inward normal
```
mgeier: sfs/array.py:512:76: W291 trailing whitespace

sfs/array.py
```diff
@@ -518,5 +621,25 @@ def _linear_helper(ycoordinates, center, orientation):
 
 
 def concatenate(*arrays):
-    """Concatenate `ArrayData` objects."""
+    """Concatenate `ArrayData` objects.
+
+    Returns
+    -------
+    `ArrayData`
+        Positions, orientations and weights of the concatenated secondary sources.
```
mgeier: sfs/array.py:629:80: E501 line too long (82 > 79 characters)

In this case, the limit should actually be 72, see https://www.python.org/dev/peps/pep-0008/#maximum-line-length.

mgeier:
> please double check the bugfix in planar() where N1 and N2 were exchanged

AFAICT this was not a bug.

The words "horizontal" and "vertical" in the docstring refer to the array's  default `orientation=[1, 0, 0]`.

With your "fix" the behavior is reversed with regards to the description.

fs446:
> > please double check the bugfix in planar() where N1 and N2 were exchanged
> 
> AFAICT this was not a bug.
> 
> The words "horizontal" and "vertical" in the docstring refer to the array's default `orientation=[1, 0, 0]`.
> 
> With your "fix" the behavior is reversed with regards to the description.

I double checked what's going on. In fact neither the original version as well as the "bugfix" solving the problem in a consistent way, both sometimes are correct and sometimes not in terms of math. For orientation = (0,0,1) the fixed version is doing what the user might expect from the given documentation, that is the current example. However, we need to fix this for proper and general handling. I'd suggest:
In order to define a plane discretized with du and dv we could define this plane with the two orthogonal! vectors u and v (u dot v = 0) and calculate the resulting normal vector n = u cross v. N1 and N2 then belong to u and v, respectively and consistently should be renamed to Nu and Nv (or whatever two letters we use later for denotation). The function call then might read as

planar((u,v), (Nu,Nv), (du,dv), center=[0, 0, 0])

which is mathematically consistent. Problem here is that u and v must be orthogonal, since remaining calculations rely on that assumption. However, this appears to be the best handling rather than others that involve additional rotation/looking direction information when using only u.

In order for simple and fast usage we should think about very often used defaults, such as 'xy', 'yz', 'zx' planes with normals (i) into the remaining cartesian axis or (ii) flipped (equivalent to 'yx', 'zy', 'xz'-planes). I don't know what the most elegant handling would be.

sfs/array.py
```diff
@@ -422,39 +459,97 @@ def cube(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
...
+    Examples
+    --------
+    content of `example_array_6LS_3D.txt` 
+    ::
+        1 0 0 1
+        -1 0 0 1
+        0 1 0 1
+        0 -1 0 1
+        0 0 1 1
+        0 0 -1 1
```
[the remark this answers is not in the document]
fs446: Retyping is on purpose. There might be cases when the docu or the loaded text file is missing, and then the load() function still indicates an example handling. I'd suggest to keep this.

sfs/array.py
```diff
@@ -422,39 +459,97 @@ def cube(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
...
         Positions, orientations and weights of secondary sources.
 
+    Examples
+    --------
+    content of `example_array_4LS_2D.csv`
+    ::
+        1,0,0,-1,0,0,1
+        0,1,0,0,-1,0,1
+        -1,0,0,1,0,0,1
+        0,-1,0,0,1,0,1
```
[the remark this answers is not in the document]
fs446: Retyping is on purpose. There might be cases when the docu or the loaded text file is missing, and then the load() function still indicates an example handling. I'd suggest to keep this.

sfs/array.py
```diff
@@ -484,6 +579,14 @@ def weights_midpoint(positions, closed):
     (N,) numpy.ndarray
         Weights of secondary sources.
 
+    Examples
+    --------
+    >>> import sfs
+    >>> x0, n0, a0 = sfs.array.circular(2**5, 1)
+    >>> a = sfs.array.weights_midpoint(x0, closed=True)
+    >>> print(max(abs(a0-a)))
+    0.0003152601902411123
```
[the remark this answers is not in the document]
fs446: very good, thanks!

sfs/array.py
```diff
@@ -484,6 +579,14 @@ def weights_midpoint(positions, closed):
     (N,) numpy.ndarray
         Weights of secondary sources.
 
+    Examples
+    --------
+    >>> import sfs
+    >>> x0, n0, a0 = sfs.array.circular(2**5, 1)
+    >>> a = sfs.array.weights_midpoint(x0, closed=True)
+    >>> print(max(abs(a0-a)))
```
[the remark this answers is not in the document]
fs446: I'll remove print. However, I like the idea of putting the code into a non-interactive python shell (with of course removing >>>) and obtaining the intended return values.

sfs/array.py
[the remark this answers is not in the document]
fs446: fixed

fs446: fixed.
Thanks for suggesting, I usually use PyCharm, but not on these corrections. Enabled this option now.

sfs/array.py
fs446: fixed

sfs/array.py
fs446: all changed to return

sfs/array.py
fs446: changed to handling that is PEP8 ok

mgeier:
@fs446 Can you please remove your "fix" from this PR and create a new issue (and/or PR if you want) for the issue you described above?

fs446:
could do this, but I'd vote for 
https://github.com/sfstoolbox/sfs-python/commit/de2942581d782aae8d472a65768e8789ed1c966a

sfs/array.py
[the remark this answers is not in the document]
mgeier: I very much dislike teaching people to write things that are utterly unnecessary and make their life harder.

mgeier:
Now that `pycodestyle` is satisfied, the next step is:

```
$ python3 -m pydocstyle sfs/array.py 
[...]
sfs/array.py:244 in public function `rounded_edge`:
        D205: 1 blank line required between summary line and description (found 0)
sfs/array.py:244 in public function `rounded_edge`:
        D400: First line should end with a period (not 'd')
sfs/array.py:319 in public function `edge`:
        D205: 1 blank line required between summary line and description (found 0)
sfs/array.py:319 in public function `edge`:
        D400: First line should end with a period (not 'p')

**Comment by mgeier** [comment]
@fs446:

> could do this, but I'd vote for https://github.com/sfstoolbox/sfs-python/commit/de2942581d782aae8d472a65768e8789ed1c966a

I'd vote against, because this doesn't make sense given the current text in the docstring:

> If a pair of numbers is given, the first one specifies the number on the horizontal edge, the second one specifies the number on the vertical edge.

If `orientation=[0, 0, 1]` is given, there is no vertical edge, there are only two horizontal edges!

I guess we have different understanding of the words "orientation" and/or "edge" and/or "vertical" and/or "horizontal". Or probably "x", "y" and "z"?

Just to be clear: "horizontal" means for me: any direction that has a zero z-component, i.e. `[x, y, 0]`. "vertical" means for me: the direction that has only a z-component, i.e. `[0, 0, z]`.

Linear arrays are by default "horizontal", so are rectangular and circular ones (which I guess is the typical use case).

The `orientation` in the `sfs.array` module so far has meant the orientation of the loudspeakers themselves (or one reference loudspeaker in case they have different orientations), not necessarily the lines or planes or whatever they are on.

Planar arrays should IMHO have a default primary radiation direction that is horizontal (which is the main thing I've seen on real loudspeaker arrays so far).
Your suggestion `orientation=[0, 0, 1]` would mean that the main radiation direction would be up to the ceiling.

I agree that there is a problem with the orientation of arrays (in contrast to assumed rotational-symmetric-along-their-main-axis single loudspeakers) because when only providing one orientation vector, one degree of freedom stays undefined. That's why I opened #36.
Currently, the rotations are still well-defined (except probably `orientation=[-1, 0, 0]`?), but the chosen rotation might not be obvious.
And what's probably worse, it's not possible to get to all possible rotations.
You might have to do a second (and third?) rotation to get to an arbitrary orientation.

> both sometimes are correct and sometimes not in terms of math.

For the version before your changes, do you mean the case `orientation=[-1, 0, 0]`?
Or some other cases?

> planar((u,v), (Nu,Nv), (du,dv), center=[0, 0, 0])

This sounds overly complicated, but more importantly it's not consistent with the current API of `linear()` and the other functions in the module.
How would you extend those to become consistent?

> In order for simple and fast usage we should think about very often used defaults, such as 'xy', 'yz', 'zx' planes with normals (i) into the remaining cartesian axis or (ii) flipped (equivalent to 'yx', 'zy', 'xz'-planes). I don't know what the most elegant handling would be.

Probably it would be enough to provide a list of example invocations in the documentation?

**Comment by fs446** [comment]
> 
> On 7. Mar 2019, at 3:58 PM, Matthias Geier <notifications@github.com> wrote:
> 
> @fs446:
> 
> could do this, but I'd vote for de29425
> 
> I'd vote against, because this doesn't make sense given the current text in the docstring:
> 
> If a pair of numbers is given, the first one specifies the number on the horizontal edge, the second one specifies the number on the vertical edge.
> 
> If orientation=[0, 0, 1] is given, there is no vertical edge, there are only two horizontal edges!
> 
> I guess we have different understanding of the words "orientation" and/or "edge" and/or "vertical" and/or "horizontal". Or probably "x", "y" and "z"?
> 
> Just to be clear: "horizontal" means for me: any direction that has a zero z-component, i.e. [x, y, 0]. "vertical" means for me: the direction that has only a z-component, i.e. [0, 0, z].
> 
> Linear arrays are by default "horizontal", so are rectangular and circular ones (which I guess is the typical use case).
> 
> The orientation in the sfs.array module so far has meant the orientation of the loudspeakers themselves (or one reference loudspeaker in case they have different orientations), not necessarily the lines or planes or whatever they are on.
> 
> Planar arrays should IMHO have a default primary radiation direction that is horizontal (which is the main thing I've seen on real loudspeaker arrays so far).
> Your suggestion orientation=[0, 0, 1] would mean that the main radiation direction would be up to the ceiling.
> 
Ok, now I see your point! Slightly different viewpoints/understandings indeed, but yours is consistent with the current usage. I will restore to state before "bugfixing".

> I agree that there is a problem with the orientation of arrays (in contrast to assumed rotational-symmetric-along-their-main-axis single loudspeakers) because when only providing one orientation vector, one degree of freedom stays undefined. That's why I opened #36.
Yes, that is about the same problem then.

> Currently, the rotations are still well-defined (except probably orientation=[-1, 0, 0]?), but the chosen rotation might not be obvious.
> And what's probably worse, it's not possible to get to all possible rotations.
> You might have to do a second (and third?) rotation to get to an arbitrary orientation.
> 
> both sometimes are correct and sometimes not in terms of math.
> 
> For the version before your changes, do you mean the case orientation=[-1, 0, 0]?
Yes.
> Or some other cases?
> 
> planar((u,v), (Nu,Nv), (du,dv), center=[0, 0, 0])
> 
> This sounds overly complicated,
For me its not, but probably a user should not be forced to span a vector space just to define a planar SSD.
The handling would be most consistent in terms of math.
> but more importantly it's not consistent with the current API of linear() and the other functions in the module.
Yeah, I was not thinking about the consequences.
> How would you extend those to become consistent?
> 
Don't know yet.

> In order for simple and fast usage we should think about very often used defaults, such as 'xy', 'yz', 'zx' planes with normals (i) into the remaining cartesian axis or (ii) flipped (equivalent to 'yx', 'zy', 'xz'-planes). I don't know what the most elegant handling would be.
> 
> Probably it would be enough to provide a list of example invocations in the documentation?
> 
Yes, that's a good idea. I will provide this.

**Comment by fs446** [comment]
ready for master?

**Comment by mgeier** [comment]
We are very close.

This is still not addressed: https://github.com/sfstoolbox/sfs-python/pull/97#issuecomment-470542347.

File: sfs/array.py [hunk 24]
```diff
@@ -34,6 +34,7 @@ class SecondarySourceDistribution(namedtuple('SecondarySourceDistribution',
     __slots__ = ()
 
     def __repr__(self):
+        """Return representation string."""
```
mgeier: I personally find that this docstring is utterly superfluous.

Sadly, `pydocstyle` recommends it, see https://github.com/PyCQA/pydocstyle/issues/60.

sfs/array.py
```diff
@@ -343,7 +359,8 @@ def rounded_edge(Nxy, Nr, dx, center=[0, 0, 0], orientation=[1, 0, 0]):
 
 
 def edge(Nxy, dx, center=[0, 0, 0], orientation=[1, 0, 0]):
-    """Array along the xy-axis with edge at the origin.
+    """
+    Return SSD along the xy-axis with sharp edge at the origin.
```
mgeier: A leading `\n` is allowed per PEP 257, but none of our docstrings use it.

PR #99, opened 2019-02-28
Add time-domain examples in docstring

narahahn:


sfs/time/drivingfunction.py
```diff
@@ -2,6 +2,34 @@
...
+
+    plt.rcParams['figure.figsize'] = 6, 6
+
+    xs = -1.5, 1.5, 0  # point source
+    npw = sfs.util.direction_vector(np.radians(-45))  # plane wave
+    xf = -0.5, 0.5, 0  # focused source
+    nf = sfs.util.direction_vector(np.radians(-45))  # normal vector for focused source
+
+    # Impulsive excitation
+    signal = sfs.util.DelayedSignal(np.insert(np.zeros(511), 0, 1), 44100, 0)
```
mgeier: This is not at all how this is meant to be used!

As the documentation (http://python.sfstoolbox.org/en/0.4.0/utilities.html#sfs.util.DelayedSignal) says:

> This class (a `collections.namedtuple`) is not meant to be instantiated by users.
>
> To pass a signal to a function, just use a simple `tuple` or `list` containing the audio data and the sampling rate (in Hertz), with an optional starting time (in seconds) as a third item.

The future goal for using this is like that:

```python
signal = [1], 44100
```

... but sadly, this doesn't work yet because the resulting signals are cropped. We still have to come up with a solution to that.

For now, it should suffice to add a few zeros:

```python
signal = [1, 0, 0, 0], 44100
```

Do you really need that many zeros?

You can of course also use a NumPy array, but you should not use `DelayedSignal`.

sfs/time/source.py
```diff
@@ -4,6 +4,23 @@
...
+    import numpy as np
+    import sfs
+
+    plt.rcParams['figure.figsize'] = 6, 6  # inches
+
+    xs = 1.5, 1, 0  # point source
+    t = np.linalg.norm(xs) / sfs.default.c
+
+    # Impulsive excitation
+    signal = sfs.util.DelayedSignal(np.insert(np.zeros(511), 0, 1), 44100, 0)
```
mgeier: Same as above.

sfs/time/source.py
```diff
@@ -4,6 +4,23 @@
...
+    :context: reset
+
+    import matplotlib.pyplot as plt
+    import numpy as np
+    import sfs
+
+    plt.rcParams['figure.figsize'] = 6, 6  # inches
+
+    xs = 1.5, 1, 0  # point source
+    t = np.linalg.norm(xs) / sfs.default.c
```
mgeier: What's the idea behind choosing this time instant?

I understand that it's the time when the wave front crosses the origin, but it looks more confusing than necessary to me.

What about providing a "distance travelled by the sound wave" in meters and then converting that into the observation time in seconds?

sfs/time/source.py
```diff
@@ -4,6 +4,23 @@
...
 .. include:: math-definitions.rst
 
+.. plot::
+    :context: reset
+
+    import matplotlib.pyplot as plt
+    import numpy as np
+    import sfs
+
+    plt.rcParams['figure.figsize'] = 6, 6  # inches
```
mgeier: This doesn't work too great with the second plot, because it's not quadratic.

What about removing this line and using `figsize=[6, 6]` for creating the quadratic plot?

Or alternatively, what about making all the plots (i.e. both plots) more or less the same shape?

mgeier: #100 should improve this a bit, but it's still a good idea to use a `figsize` that is appropriate for the shape of the plot.

sfs/time/drivingfunction.py
narahahn: `as_delayed_signal` shouldn't be used, either?

We need more zeros for NFC-HOA driving signals.

sfs/time/source.py
narahahn: OK, no problem.

sfs/time/source.py
narahahn: Now the examples in `time.source` are plotted in rectangles,
the same as in `mono.source`.

I think squares look better for SFS examples in `time.drivingfunction`,
as we are using a circular array.

sfs/time/drivingfunction.py
mgeier: > `as_delayed_signal` shouldn't be used, either?

Not in this case, no.
If you create a signal in a script, it is sufficient and preferable to simply use a `tuple` (or a `list`).

The helper function `as_delayed_signal()` is meant to be used inside a function that takes a signal as an input.

Basically every function in `sfs.time` that takes a `signal` argument will use `as_delayed_signal()` (unless it directly forwards the `signal` argument to another function which in turn will use `as_delayed_signal()`).

> We need more zeros for NFC-HOA driving signals.

OK, then it makes sense to use `np.zeros()` at some point. I find the `np.insert()` thing a bit confusing, but I guess it's OK.

narahahn: The unit impulse is now generated by [`scipy.signal.unit_impulse()`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.unit_impulse.html).

mgeier: Cool, I didn't even know such a function existed, very handy!

sfs/time/source.py
mgeier: I took the liberty of using `unit_impulse()` in bf104a3dae8b1ff44cf91fa8bdd3010daff2b19f.

PR #101, opened 2019-03-02
Introduce synthesize() functions for soundfield superposition

mgeier:
* change `sfs.mono.synthesized.generic()` to `sfs.mono.synthesize()`
* change `sfs.time.soundfield.p_array()` to `sfs.time.synthesize()`
* "driving functions" now return "secondary source functions" in addition to what they returned before
  * those functions incorporate information about `c` and `omega`
    * therefore, `c` and `omega` don't have to be passed to the "synthesize" function
  * those functions are supposed to be passed on to `synthesize()`
    * therefore, "source" functions don't have to accept a normal vector anymore if it is unused
    * if desired, a user can still provide a different functions with the required signature
* rename `sfs.array.ArrayData` to `sfs.array.SecondarySourceDistribution`

Those changes are closely related but still to some extent independent. We could merge only a subset of them, if desired.

All this is of course very much open for discussion.

What I don't like in `sfs.mono.synthesize()` is that sometimes the "weight" is not needed and has to be manually set to all ones. I guess this happens mostly for NFC-HOA, because there is no source selection happening ...

I didn't change those driving functions which don't appear in any of the examples and have no example use in their docstring. Those should be changed by somebody other than me in a separate PR in case this PR is merged.


fs446:
Nice, thanks. I just checked frequency domain, which works as expected for my tested cases.

>Those changes are closely related but still to some extent independent. We could merge only a subset of them, if desired.

I'd accept all changes. They are very meaningful to me and simplify workflow.

I'd change the order of variables `def synthesize(signals, weights, secondary_source, array, **kwargs)` which appears more logical to me in terms of the processing and signal flow.

I'd suggest to avoid naming `array` to link the secondary_source_distribution and secondary_source closer together making reading more intuitive. Since the class is already called SecondarySourceDistribution which is very good, I don't know what best practice is, but I'd vote for exchanging `array` by `secondary_source_distribution`.

I don't like the usage of wording 'weights' that much. At least we should avoid the double use for the integration weights stored in array and some other weights steming from tapering and sec source selection. I think it would help to identify the array weights more clearly as 'integration weights' in the docstring help or find a completely different name. This reduces misconception for beginners a lot I guess.

>What I don't like in sfs.mono.synthesize() is that sometimes the "weight" is not needed and has to be manually set to all ones. I guess this happens mostly for NFC-HOA, because there is no source selection happening ...

That's unfortunate, but I could live with it. However, one workaround could be: Your suggested handling now allows to calculate the secondary selection within the driving_function function.  We could return this from the function as well. HOA driving functions then would return unity gain array, plain VBAP or WFS return the appropriate secondary source selection. Same as you mentioned for secondary_source handling: we can use it, but we don't have to for calling synthesize().

>I didn't change those driving functions which don't appear in any of the examples and have no example use in their docstring. Those should be changed by somebody other than me in a separate PR in case this PR is merged.

I can do this, once we agreed on taking this effort.

mgeier:
After some offline discussions with @fs446 I made a few changes (8bbe7c79f413cf620697b37fdcd9404010a58e7f):

* incorporate the calculation of the secondary source selection into the driving functions
  * this can be used to calculate tapering (if desired) and the resulting weights can be passed to `synthesize()`
  * if desired, a non-default source selection can still be used by just ignoring the returned source selection
* change the order of parameters in `synthesize()`
* I had to introduce an additional parameter `ns` for the focused driving functions in order to be able to return the source selection.

Please note that the NFC-HOA functions also return the secondary source selection, even though all loudspeakers are always active.

fs446:
Thanks @mgeier!, I appreciate this effort very much. Checked some driving functions with a test jupyter notebook for frequency domain, works as expected. Also DocString Examples work. So I guess behaves as desired at the moment.

The handling appears very logic to me: (i) calling the driving function, getting all relevant and specific information back from drivingfunction, (ii) add tapering by taste manually and then (iii) calling synthesize() with all gathered information. I don't see any flaws of this approach at the moment. Furthermore, we still have all degrees of freedom to change the secondary source function or the secondary selection, just by not using what driving function returns, but defining customised variable/function pointer that gets into synthesize().

Providing HOA driving functions with a secondary source selection is IMO no category error, this information is rather a primary & secondary source and sound field synthesis technique dependent parameter and should aways available, even if it appears useless for NFC-HOA at the moment. Think of an VBAP implementation for one specific direction in this style, then you would have exactly three sources selected with True, remaining are non-selected with False. It is consistent, but also could be seen useless. However, maybe fancy new Ambisonics or other approaches coming up one day where this selection tool included might be meaningful. IMO it is definitely not a WFS specific thing as we utilised it so far.

I'd vote to adopt this to the master branch very soon in order to move forward with the driving function and plotting changes.

fs446:
After forced push to d95e371  I just checked WFS time domain, works also as expected.

mgeier:
Thanks for checking @fs446. I just combined two commits, so there was no logical change anyway.

I'm just waiting for the @narahahn's review, then we can merge!

sfs/mono/__init__.py
```diff
@@ -5,11 +5,56 @@
...
+from .. import array as _array
+
+
+def shiftphase(p, phase):
+    """Shift phase of a sound field."""
+    p = _np.asarray(p)
+    return p * _np.exp(1j * phase)
+
+
+def synthesize(d, weights, ssd, secondary_source_function, **kwargs):
```
narahahn: The variable name `array` is mostly used in other codes.
Is there a reason using `ssd` here?

sfs/time/__init__.py
```diff
@@ -5,9 +5,61 @@
...
 """
 from . import drivingfunction
 from . import source
-from . import soundfield
+
+from .. import util as _util
+from .. import array as _array
+
+
+def synthesize(signals, weights, ssd, secondary_source_function, **kwargs):
```
narahahn: `ssd` or `array`?

narahahn:
I just commented about consistent naming, either `ssd` vs `array`.

Other than that, I think it can be merged.

I'll prepare the take over the changes to the branches I'm working on (#69 and #45).

mgeier:
@fs446 Thanks for merging!

@narahahn We still have to find the right balance between `ssd` and `array`. For now I've used `array` in code examples, while most of our functions use `x0`, `n0`, `a0` or, since recently, `ssd`.

I'm not sure if the module `sfs.array` is OK or if we should rename it.

If you feel something should be renamed, please make an issue or a PR.

PR #103, opened 2019-03-03
Add new equation link examples to sfs.rtfd.io

hagenw:
This is a follow up on #53 and https://github.com/sfstoolbox/theory/pull/29 and meant to merge at its current state.
It shows with the example of the driving functions how it might look like.

I adjusted the naming to use `freq` and `time`, but I'm still not sure if it might look better to have something like `frequency-domain` and `time-domain` at the end of the links instead.

In addition, I figured out that we can easily configure the main URL to the theory using `extlinks`. This is nice as it allows for easy switching between different versions of the documentation. Currently I link to https://sfs.readthedocs.io/en/v3-equation-names which is a special branch I created to test this pull request.

@mgeier let me know proposals for better naming schemes. If I remember correctly, you didn't liked the short `d_wfs` part at the beginning. I stayed with it as the naming `wfs` and `drivingfunction` is repeated in the actual URL name and to avoid to long URLs.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -40,6 +40,7 @@ extensions = [
     'sphinx.ext.napoleon',  # support for NumPy-style docstrings
     'sphinx.ext.intersphinx',
     'sphinx.ext.doctest',
+    'sphinx.ext.extlinks',
     'sphinxcontrib.bibtex',
     'matplotlib.sphinxext.plot_directive',
     'nbsphinx',
@@ -76,6 +77,9 @@ intersphinx_mapping = {
     'scipy': ('https://docs.scipy.org/doc/scipy/reference/', None),
     'matplotlib': ('http://matplotlib.org/', None),
 }
+
+extlinks = {'sfs': ('https://sfs.readthedocs.io/en/v3-equation-names/%s',
+                    'https://sfs.rtfd.io/')}
 
 plot_include_source = True
 plot_html_show_source_link = False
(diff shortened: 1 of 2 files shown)


mgeier:
The naming situation in SFS-Python is still very much in flux. For example, I'm still hoping to get rid of the string "drivingfunction", which is much too verbose for my taste.

> If I remember correctly, you didn't liked the short `d_wfs` part at the beginning.

Yeah, since equation IDs are unique, any additional part of the URI is redundant.

I think it would be nice to have a Sphinx extension to handle the equation IDs, but AFAIK such a thing doesn't exist yet (see https://github.com/sfstoolbox/theory/pull/29#issuecomment-382468365).

Or did you try something like this @hagenw?

I don't think I currently have the time to look into this.

My initial idea was to try to make the equation IDs *identical* to the corresponding Python functions.
Do you think this is a reasonable goal?

Some theoretical examples (pending some re-naming in SFS-Python):

    https://sfs.rtfd.io/#sfs.time.wfs.point_25d
    https://sfs.rtfd.io/#sfs.util.source_selection_point

One problem is that there are certainly many equations that don't have a Python implementation (yet).

Another problem is that whenever we change a function/module name in Python it would become disconnected from the Theory ID.

hagenw:
Yes, I remember our discussion that the equation ID would be sufficient for a link and I'm still hoping to get there some day. Unfortunately, I was not able to find an easy solution and don't have the time at the moment to come up with something by myself.

For that reason I decided to start with something as my goal is to have one version up and running before DAGA. As the versioning is done probably now it isn't a big deal if we change the URL later on as the old ones will still be available.

I understood also the goal of identical naming, but it will not be possible all the time. For example, in the theory of WFS there is no meaningful distinction between 2D and 3D and the driving function is valid in both cases as in the toolbox we distinguish between 2D and 3D for WFS.
But again, we can also update this in a next version. At the moment I'm looking for the best compromise that can be implemented until next week.

hagenw:
OK, I will try to match them as best as possible and update this pull request afterwards. In addition, I also have to add the links for equations outside of `sfs/mono/drivingfunction.py`.

hagenw:
With #152 this is now obsolete.

PR #104, opened 2019-03-06
Documentation of driving functions

spors:
- Enhanced existing documentation of driving functions
- Added documentation to various driving functions
- New examples for SDM approaches

What should we do about...
- documention of `wfs_3d_point` and `wfs_2d_point` is the same due to their definitions. The examples are a bit misleading in this context. Should we define seperate functions for 2/3-dimensional case? The same holds for `wfs_2d_plane` and `wfs_2d_focused`
- the examples for SDM redefine the circular array to a linear one. Is this OK?

### DIFF
--- a/sfs/mono/drivingfunction.py
+++ b/sfs/mono/drivingfunction.py
@@ -40,8 +40,31 @@ from .. import default
 
 
 def wfs_2d_line(omega, x0, n0, xs, c=None):
-    r"""Line source by 2-dimensional WFS.
+    r"""Driving function for 2-dimensional WFS for a virtual line source.
 
+    Parameters
+    ----------
+    omega : float
+        Frequency of line source.
+    x0 : (N, 3) array_like
+        Sequence of secondary source positions.
+    n0 : (N, 3) array_like
+        Sequence of normal vectors of secondary sources.
+    xs : (3,) array_like
+        Position of virtual line source.
+    c : float, optional
+        Speed of sound.
+
+    Returns
+    -------
+    (N,) numpy.ndarray
+        Complex weights of secondary sources.
+
+    Notes
+    -----
+    .. math::
+
+        V(\x,\w) = \frac{1}{\rho c} \e{-\i\wc\n\x} \n
     .. math::
 
         D(\x_0,\w) = \frac{\i}{2} \wc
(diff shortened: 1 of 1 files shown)


fs446:
> * documention of `wfs_3d_point` and `wfs_2d_point` is the same due to their definitions. The examples are a bit misleading in this context. Should we define seperate functions for 2/3-dimensional case? The same holds for `wfs_2d_plane` and `wfs_2d_focused`

`wfs_3d_point` is point source for a 3D SSD, right? So a planar SSD would be meaningful for that.
`wfs_2d_point` seems to be odd, what is the purpose of that driving function, synth a 3D point source with a 2D line source SSD, this is not going to work? `wfs_2d_plane` works in 3D and 2D as long in 2D the PW direction is restricted to 2D. `wfs_2d_focused` seems again odd to me, but surely I'm just missing somewhat here. Anyway, even if we have double cases, I'd always make the complete help, maybe giving a hint to a copy/clone/coincidences.

> the examples for SDM redefine the circular array to a linear one. Is this OK?
Totally ok for me and of course also necessary to show a meaningful example. Same holds for the ESA driving function using an edged SSD, are you going to add these examples as well?

mgeier:
> documention of `wfs_3d_point` and `wfs_2d_point` is the same due to their definitions.

Instead of two different names for one function, we could make two different functions (with their own docstrings) that call the same (hidden) function. This way we only duplicate parts of the docstring, but not the actual code.

It's not even strictly necessary to have a hidden function with the actual implementation. We could just implement one function and then make another function that calls the first one.
But I think having a hidden function that contains the implementation is easier to follow.

spors:
Thanks for the discussion. I will leave the hidden functions as they are and call it from the actual functions with different docstrings.

sfs/mono/drivingfunction.py
```diff
@@ -265,7 +415,32 @@ def wfs_25d_focused(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
...
+    -------
+    float
+        Complex weight for given angular frequency.
+
+    Notes
+    -----
+    .. math::
+
+        H(\w) = \begin{cases}
+            \sqrt{\i \wc} & \text{for } \w <= \w_\text{alias} \\
```
mgeier: `<=` -> `\leq`

fs446:
I started to rebase this stuff with solving conflicts in the [documentation2merge](https://github.com/sfstoolbox/sfs-python/tree/documentation2merge) branch. 

Then I added SDM and ESA examples into docstrings. ESA Dipoles are still missing, this secondary source is not yet defined. Some small code modifications were necessary, please double check them.

Return values for docstring is proposed in sfs.mono.drivingfunction.wfs_2d_line()

fs446:
these modifications were handled in 3b4e0b6ce0ca0c507e1313c90491b6ea86938171, thus closing this PR

Issue #25, opened 2016-12-11
Rename frequency domain and time domain submodules


mgeier:
Currently they are called `mono` and `time`, which is inconsistent.

Possible options:

* `frequency` + `time`
* `fd` + `td`
* `f` + `t`

Another option, which is currently used in the equation references on http://sfstoolbox.org/, is to use capital letters to denote the frequency domain.

Any other ideas?

chris-hld:
Additionally, there are some functions, that are categorized in either, but are actually valid for both domains.
E.g. `sfs.mono.drivingfunction.source_selection_...` comes to my mind.

spors:
I would vote for fd for frequency-domain and td for time-domain functions. However at some point we might also have to distinguish between temporal and spatial frequency domain. For instance when implementing the spectral divison method (SDM).

trettberg:
> at some point we might also have to distinguish between temporal and spatial frequency domain

Maybe we could consistently use "modal domain" for the various spatial frequency domains.

hagenw:
I'm also in favor of "time domain", "frequency domain", and "modal domain", but I'm not completely sure if it is sufficient to use `td`, `fd`, `md`, or if those names should be more self-explanatory to the user.

hagenw:
There are two things I like about the current (`time`, `mono`) solution:
1.) They are short and have equal length.
2.) You can pronounce them.

As @spors has pointed out `frequency` would have a problem. It might be that the usage of `time` and `mono` is not completely consistent, but they are not as problematic.
`modal` as a third option would be fine with me.

fietew:
I think, that ``modal`` is too general. In the future, we might need a namespace for each of the spatial frequency domains, e.g. ``cht``, ``sht``, ``kx`` or ``pwd``.

fietew:
I do not see a problem with using ``freq`` for the time-frequency domain, since we do not have the (one) space-frequency. Most likely, we will refer to the specific geometry-dependent domains ``cht``, ``sht``, ``kx`` or ``pwd``, instead. Right now, I do not see any functionality we can implement without knowing the specific space-frequency domain. Hence, ``modal`` is too general, imho.

We should also consider, that ``modal`` (or the specific domain) has to be likely combined with both, ``time`` and  ``freq``. I think, that an additional level of hierarchy e.g. ``time.cht`` and ``freq.cht`` might be better suited for that.

hagenw:
[...]

PR #106, opened 2019-03-06
Move source selection helpers to sfs.util

mgeier:
They are not specific to monochromatic applications, as @chris-hld already mentioned in https://github.com/sfstoolbox/sfs-python/issues/25#issuecomment-266313599.

I didn't know any better place than `sfs.util`, but if somebody else does, please speak up!


fs446:
appears to be the best place for the upcoming changes, go ahead.

spors:
I would also support this!

mgeier:
OK, thanks, I've merged it.

PR #110, opened 2019-03-11
DOC: Escape underscore in \text{}

mgeier:
Unescaped underscores seem to work with MathJax, but LaTeX itself does complain.


hagenw:
Good idea.

BTW, in KaTeX you also have to use the underscore.

PR #113, opened 2019-03-11
Copyright year update

mgeier:


hagenw:
In `doc/conf.py` you could do the copyright statement like this:
```python
from datetime import datetime
copyright = '{}, {}'.format(datetime.now().year, author)
```
Then you will no longer need to update it manually.

mgeier:
@hagenw I don't think the copyright year should depend on when somebody happens to create the docs.

To avoid this problem, we could get the year from Git, similar to what I did there:

https://github.com/sfstoolbox/sfs-python/blob/1af1cbdf7987bd3c1f1d93756faf83283dc205b9/doc/conf.py#L134-L138

... but this means if Git is not available we have no valid copyright information.

This seems brittle and not worth the effort to me.

But if you like to pursue this, please feel free to make a PR.

PR #117, opened 2019-03-11
WFS 2.5D point source

fs446:
I remember we had this discussion a while ago in the Matlab toolbox. I try it here again :-)
I'd vote to change our default WFS 2.5D point source to that driving function originating from Delft, which is compatible with the recent works of Firtha, i.e. the [unified WFS framework](https://github.com/gfirtha/gfirtha_phd_thesis).
Our currently used version is -- due to the further approximation -- not easy to handle in terms of
- consistent and proper amplitude normalization
- introducing/defining other referencing schemes

The PR is merge prepared in terms of our latest changes w.r.t. driving function and docstring example handling, but not for literature references.

I would not mind to supply both driving functions with a consistent and proper name scheme. I did not come up with a meaningful idea, though.

### DIFF
--- a/sfs/mono/wfs.py
+++ b/sfs/mono/wfs.py
@@ -155,7 +155,101 @@ point_2d = _point
 
 
 def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
+    r"""Driving function for 2.5-dimensional WFS of a virtual point source.
+
+    .. versionchanged:: 0.5
+        see notes, old handling of `point_25d()` is now `point_25d_legacy()`
+
+    Parameters
+    ----------
+    omega : float
+        Angular frequency of point source.
+    x0 : (N, 3) array_like
+        Sequence of secondary source positions.
+    n0 : (N, 3) array_like
+        Sequence of normal vectors of secondary sources.
+    xs : (3,) array_like
+        Position of virtual point source.
+    xref : (3,) array_like, optional
+        Reference point xref or contour xref(x0) for amplitude correct
+        synthesis.
+    c : float, optional
+        Speed of sound in m/s.
+    omalias: float, optional
+        Angular frequency where spatial aliasing becomes prominent.
+
+    Returns
+    -------
+    d : (N,) numpy.ndarray
+        Complex weights of secondary sources.
+    selection : (N,) numpy.ndarray
+        Boolean array containing ``True`` or ``False`` depending on
+        whether the corresponding secondary source is "active" or not.
+    secondary_source_function : callable
+        A function that can be used to create the sound field of a
+        single secondary source.  See `sfs.mono.synthesize()`.
+
+    Notes
+    -----
+    `point_25d()` derives 2.5D WFS from the 3D
+    Neumann-Rayleigh integral (i.e. the TU Delft approach).
+    The eq. (3.10), (3.11) in :cite:`Start1997`, equivalent to
+    Eq. (2.137) in :cite:`Schultz2016`
+
+    .. math::
+
+        D(\x_0,\w) = \sqrt{8 \pi \, \i\wc}
+            \sqrt{\frac{|\x_\text{ref}-\x_0| \cdot
+            |\x_0-\x_\text{s}|}{|\x_\text{ref}-\x_0| + |\x_0-\x_\text{s}|}}
+            \scalarprod{\frac{\x_0-\x_\text{s}}{|\x_0-\x_\text{s}|}}{\n_0}
+            \frac{\e{-\i\wc |\x_0-\x_\text{s}|}}{4\pi\,|\x_0-\x_\text{s}|}
+
+    is implemented.
+    The theoretical link of `point_25d()` and `point_25d_legacy()` was
+    introduced as *unified WFS framework* in :cite:`Firtha2017`.
+
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        d, selection, secondary_source = sfs.mono.wfs.point_25d(
+            omega, array.x, array.n, xs)
+        normalize_gain = 4 * np.pi * np.linalg.norm(xs)
+        plot(normalize_gain * d, selection, secondary_source)
+
+    """
+    x0 = util.asarray_of_rows(x0)
+    n0 = util.asarray_of_rows(n0)
+    xs = util.asarray_1d(xs)
+    xref = util.asarray_1d(xref)
+    k = util.wavenumber(omega, c)
+
+    ds = x0 - xs
+    dr = xref - x0
+    s = np.linalg.norm(ds, axis=1)
+    r = np.linalg.norm(dr, axis=1)
+
+    d = (
+        preeq_25d(omega, omalias, c) *
+        np.sqrt(8 * np.pi) *
+        np.sqrt((r * s) / (r + s)) *
+        inner1d(n0, ds) / s *
+        np.exp(-1j * k * s) / (4 * np.pi * s))
+    selection = util.source_selection_point(n0, x0, xs)
+    return d, selection, secondary_source_point(omega, c)
+
+
+point_3d = _point
+
+
+def point_25d_legacy(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
     r"""Driving function for 2.5-dimensional WFS for a virtual point source.
+
+    .. versionadded:: 0.5
+        `point_25d()` was renamed to `point_25d_legacy()` (and a new
+        function with the name `point_25d()` was introduced). See notes for
+        further details.
 
     Parameters
     ----------
@@ -171,6 +265,8 @@ def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
         Reference point for synthesized sound field.
     c : float, optional
         Speed of sound.
+    omalias: float, optional
+        Angular frequency where spatial aliasing becomes prominent.
 
     Returns
     -------
@@ -185,6 +281,10 @@ def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
 
     Notes
     -----
+    `point_25d_legacy()` derives 2.5D WFS from the 2D
+    Neumann-Rayleigh integral (i.e. the approach by Rabenstein & Spors), cf.
+    :cite:`Spors2008`.
+
     .. math::
 
         D(\x_0,\w) = \sqrt{\i\wc |\x_\text{ref}-\x_0|}
@@ -192,14 +292,19 @@ def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
             {|\x_0-\x_\text{s}|^\frac{3}{2}}
             \e{-\i\wc |\x_0-\x_\text{s}|}
 
+    The theoretical link of `point_25d()` and `point_25d_legacy()` was
+    introduced as *unified WFS framework* in :cite:`Firtha2017`.
+    Also cf. Eq. (2.145)-(2.147) :cite:`Schultz2016`.
+
     Examples
     --------
     .. plot::
         :context: close-figs
 
-        d, selection, secondary_source = sfs.mono.wfs.point_25d(
+        d, selection, secondary_source = sfs.mono.wfs.point_25d_legacy(
             omega, array.x, array.n, xs)
-        plot(d, selection, secondary_source)
+        normalize_gain = np.linalg.norm(xs)
+        plot(normalize_gain * d, selection, secondary_source)
 
     """
     x0 = util.asarray_of_rows(x0)
@@ -215,9 +320,6 @@ def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
         r ** (3 / 2) * np.exp(-1j * k * r))
     selection = util.source_selection_point(n0, x0, xs)
     return d, selection, secondary_source_point(omega, c)
-
-
-point_3d = _point
 
 
 def _plane(omega, x0, n0, n=[0, 1, 0], c=None):
@@ -499,7 +601,8 @@ def preeq_25d(omega, omalias, c):
 
         H(\w) = \begin{cases}
             \sqrt{\i \wc} & \text{for } \w \leq \w_\text{alias} \\
-            \sqrt{\i \frac{\w_\text{alias}}{c}} & \text{for } \w > \w_\text{alias}
+            \sqrt{\i \frac{\w_\text{alias}}{c}} &
+            \text{for } \w > \w_\text{alias}
             \end{cases}
 
     """
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -70,3 +70,29 @@
     year = {1984},
     doi = {10.1121/1.390983}
 }
+@article{Firtha2017,
+    author = {Gergely Firtha AND P{\'e}ter Fiala AND Frank Schultz AND
+        Sascha Spors},
+    title = {{Improved Referencing Schemes for 2.5D Wave Field Synthesis
+        Driving Functions}},
+    journal = {IEEE/ACM Trans. Audio Speech Language Process.},
+    volume = {25},
+    number = {5},
+    pages = {1117-1127},
+    year = {2017},
+    doi = {10.1109/TASLP.2017.2689245}
+}
+@phdthesis{Start1997,
+  author = {Evert W. Start},
+  title = {{Direct Sound Enhancement by Wave Field Synthesis}},
+  school =  {Delft University of Technology},
+  year = {1997}
+}
+@phdthesis{Schultz2016,
+    author = {Frank Schultz},
+    title = {{Sound Field Synthesis for Line Source Array Applications in
+        Large-Scale Sound Reinforcement}},
+    school = {University of Rostock},
+    year = {2016},
+    doi = {10.18453/rosdok_id00001765}
+}


sfs/mono/drivingfunction.py
```diff
@@ -169,9 +169,11 @@ def wfs_25d_point(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
     xs : (3,) array_like
         Position of virtual point source.
     xref : (3,) array_like, optional
-        Reference point for synthesized sound field.
+        Reference point xref or contour xref(x0) for amplitude correct synthesis.
```
hagenw: E501 line too long (81 > 79 characters)

sfs/mono/drivingfunction.py
```diff
@@ -186,38 +188,50 @@ def wfs_25d_point(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
 
     Notes
     -----
+    Eq. (3.10), (3.11) Start, E.W. (1997): "Direct sound enhancement by wave
+    field synthesis", doctoral thesis, TU Delft
+
+    Eq. (2.137) from Schultz, F. (2016): https://doi.org/10.18453/rosdok_id00001765
```
hagenw: E501 line too long (83 > 79 characters)

sfs/mono/drivingfunction.py
```diff
@@ -186,38 +188,50 @@ def wfs_25d_point(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
...
     Examples
     --------
     .. plot::
         :context: close-figs
 
-        d, selection, secondary_source = sfs.mono.drivingfunction.wfs_25d_point(
-            omega, array.x, array.n, xs)
-        plot(d, selection, secondary_source)
+        d, selection, secondary_source = (
+            sfs.mono.drivingfunction.wfs_25d_point_Unified_WIP(
```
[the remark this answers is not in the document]
fs446: check the final commit against master contains only I guess wfs_25d_point, previous version include the working title WIP

sfs/mono/drivingfunction.py
```diff
@@ -186,38 +188,50 @@ def wfs_25d_point(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
...
 
-        D(\x_0,\w) = \sqrt{\i\wc |\x_\text{ref}-\x_0|}
-            \frac{\scalarprod{\x_0-\x_\text{s}}{\n_0}}
-            {|\x_0-\x_\text{s}|^\frac{3}{2}}
-            \e{-\i\wc |\x_0-\x_\text{s}|}
+        D(\x_0,\w) = \sqrt{8 \pi \, \i\wc}
+            \sqrt{\frac{|\x_\text{ref}-\x_0| \cdot
+            |\x_0-\x_\text{s}|}{|\x_\text{ref}-\x_0| + |\x_0-\x_\text{s}|}}
+            \scalarprod{\frac{\x_0-\x_\text{s}}{|\x_0-\x_\text{s}|}}{\n_0}
+            \frac{\e{-\i\wc |\x_0-\x_\text{s}|}}{4\pi\,|\x_0-\x_\text{s}|}
```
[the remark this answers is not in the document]
fs446: That's a good point and holds for all preEQ approaches. I will think about a solution.

sfs/mono/drivingfunction.py
```diff
@@ -169,9 +169,11 @@ def wfs_25d_point(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
     xs : (3,) array_like
         Position of virtual point source.
     xref : (3,) array_like, optional
-        Reference point for synthesized sound field.
+        Reference point xref or contour xref(x0) for amplitude correct synthesis.
     c : float, optional
-        Speed of sound.
+        Speed of sound in m/s.
+    omalias : float, optional
+        Cut angular frequency for prefilter.
```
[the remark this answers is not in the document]
fs446: yes, that is a good idea.

fs446:
> The actual implementation of the equation looks good, I was not able to spot an error.
> 
> As I was no longer involved in actual WFS theory for some time now, I totally trust you with the decision that this is the better default driving function ;)

Thanks!
Don't wait for this issue for the next larger release. I'm not sure if/when to handle this within the next days.

sfs/mono/drivingfunction.py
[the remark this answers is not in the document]
fs446: I'd leave this for future work and changed to

omalias: float, optional
   Angular frequency where spatial aliasing becomes prominent.
such that all driving functions are consistent

sfs/mono/drivingfunction.py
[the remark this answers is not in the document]
fs446: was indeed not yet correct, fixed now

sfs/mono/drivingfunction.py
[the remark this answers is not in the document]
fs446: Left for future work, opened a PR

sfs/mono/drivingfunction.py
fs446: fixed

sfs/mono/drivingfunction.py
fs446: fixed

fs446:
I made a new push including a master rebase, so this version is up to date with latest dev.
I've added some more comments on the 3D/2D Rayleigh integral handling to give the reader more hints what's going on. I'd vote to leave the cited literature in the docstring for now to give users/readers the chance for fast access of the change.
I've introduced 'point_25d_legacy' that implements the `point_25d` of versions <=0.4.0 . This could get another name though. I've thought about something like
`def _point_25d_from_3d()`
`def _point_25d_from_2d()` (the legacy)
and having `point_25d = _point_25d_from_3d`

I also introduced a normalising gain in both the examples (which is open for all other examples as well) to make more clear that legacy is not amplitude correct at ref point, but the new one is.

fs446:
> I've thought about something like
> `def _point_25d_from_3d()`
> `def _point_25d_from_2d()` (the legacy)
> and having `point_25d = _point_25d_from_3d`

I'm not sure how to handle the docstring docu then.

mgeier:
I like the name `point_25d_legacy()`.

sfs/mono/drivingfunction.py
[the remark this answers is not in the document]
hagenw: OK, sounds good.

sfs/mono/wfs.py
```diff
@@ -154,9 +154,13 @@ def _point(omega, x0, n0, xs, c=None):
 point_2d = _point
 
 
-def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
+def point_25d_legacy(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
     r"""Driving function for 2.5-dimensional WFS for a virtual point source.
 
+    This driving function was used as standard `point_25d` in versions <=0.4.0.
```
mgeier: The word "standard" is not necessary here.

mgeier: I prefer using `` `point_25d()` `` (with parentheses).

mgeier: If you want to mention a version, you should use the `versionadded` or `versionchanged` directives, see https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html?highlight=deprecated#directive-versionadded.

In this case this could be something like

```rst
    .. versionadded:: 0.5

        This function was called `point_25d()` before. Blah blah ...
```

sfs/mono/wfs.py
```diff
@@ -185,6 +191,16 @@ def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
 
     Notes
     -----
+    `point_25d_legacy` originates from deriving 2.5D WFS from the 2D
```
mgeier: `` `point_25d_legacy()` ``

sfs/mono/drivingfunction.py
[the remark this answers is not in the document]
hagenw: OK, is handled by #143

sfs/mono/wfs.py
```diff
@@ -185,6 +191,16 @@ def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
...
+    Neumann-Rayleigh integral (i.e. the approach by Rabenstein & Spors), cf.
+    Spors, S.; Rabenstein, R.; Ahrens, J. (2008): The theory of Wave Field
+    Synthesis revisited. In: Proc. of 124th Audio Eng. Soc. Convention,
+    Amsterdam, #7358.
+    Also cf. Eq. (2.145)-(2.147) in https://doi.org/10.18453/rosdok_id00001765
```
mgeier: Please use our BibTeX-based system for references.

sfs/mono/wfs.py
```diff
@@ -185,6 +191,16 @@ def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
...
+
+    The theoretical link of `point_25d` and `point_25d_legacy` was introduced
+    as *unified WFS framework* in https://doi.org/10.1109/TASLP.2017.2689245
```
mgeier: Same as above.

sfs/mono/wfs.py
```diff
@@ -217,6 +234,92 @@ def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
     return d, selection, secondary_source_point(omega, c)
 
 
+def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
+    r"""Driving function for 2.5-dimensional WFS of a virtual point source.
+
```
mgeier: Here you should use the `versionchanged` directive.

sfs/mono/wfs.py
```diff
@@ -217,6 +234,92 @@ def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
...
+        Boolean array containing ``True`` or ``False`` depending on
+        whether the corresponding secondary source is "active" or not.
+    secondary_source_function : callable
+        A function that can be used to create the sound field of a
+        single secondary source.  See `sfs.mono.synthesize()`.
+
+    Notes
+    -----
+    In versions <=0.4.0 we used the function that is now included in
+    `point_25d_legacy`.
```
mgeier: I would move this note to the main description.

sfs/mono/wfs.py
```diff
@@ -217,6 +234,92 @@ def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
...
+
+    `point_25d` originates from deriving 2.5D WFS from the 3D
+    Neumann-Rayleigh integral (i.e. the TU Delft approach), cf. e.g.
+    Eq. (3.10), (3.11) in Start, E.W. (1997): "Direct sound enhancement by
+    wave field synthesis", doctoral thesis, TU Delft.
+    Also cf. Eq. (2.137) in https://doi.org/10.18453/rosdok_id00001765
+
+    The theoretical link of `point_25d` and `point_25d_legacy` was introduced
+    as *unified WFS framework* in https://doi.org/10.1109/TASLP.2017.2689245
```
mgeier: Again, please use BibTeX references.

sfs/mono/wfs.py
```diff
@@ -217,6 +234,92 @@ def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
...
+            |\x_0-\x_\text{s}|}{|\x_\text{ref}-\x_0| + |\x_0-\x_\text{s}|}}
+            \scalarprod{\frac{\x_0-\x_\text{s}}{|\x_0-\x_\text{s}|}}{\n_0}
+            \frac{\e{-\i\wc |\x_0-\x_\text{s}|}}{4\pi\,|\x_0-\x_\text{s}|}
+
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        d, selection, secondary_source = (
```
mgeier:
```python
        d, selection, secondary_source = sfs.mono.wfs.point_25d(
            omega, array.x, array.n, xs)
```

sfs/mono/wfs.py
```diff
@@ -217,6 +234,92 @@ def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
...
+            sfs.mono.wfs.point_25d(omega, array.x, array.n, xs))
+        normalize_gain = 4*np.pi*np.linalg.norm(xs)
```
mgeier: I personally prefer more spaces here:

```python
normalize_gain = 4 * np.pi * np.linalg.norm(xs)

**Comment by mgeier** [comment]
You should define the "legacy" function after the new function.

File: sfs/mono/wfs.py (hunk 8 above)
Comment (fs446) [reply sfs/mono/wfs.py]: Ok, I will do it. However, none of the existing driving filters/signals use this handling at the moment. Can you provide an example?
Besides that, I'd strongly suggest always adding a DOI or other permalink of the reference directly in the docstring help for users that only use the source code. This helps a lot, since we can not always guarantee that our referencing docu is correct, or can we indeed?

Comment (mgeier) [reply sfs/mono/wfs.py]: > Can you provide an example?

Sure: 

https://github.com/sfstoolbox/sfs-python/blob/a15b4f362989493597e7f381ed2d9ec23b1d08e3/sfs/mono/wfs.py#L250-L252

and

https://github.com/sfstoolbox/sfs-python/blob/a15b4f362989493597e7f381ed2d9ec23b1d08e3/doc/references.bib#L41-L46

> I'd strongly suggest always adding a DOI or other permalink of the reference directly in the docstring help for users that only use the source code.

We should either do everything manually or use a centralized approach. We should definitely not do both at the same time.
I think the centralized approach makes much more sense.

> we can not always guarantee that our referencing docu is correct, or can we indeed?

I don't know what you mean.

But if you read ``:cite:`Spors2008` `` in the source code, you can go to `doc/references.bib`, search for "Spors2008" and get all the information you need.

File: sfs/mono/wfs.py (hunk 14 above)
Comment (fs446) [reply sfs/mono/wfs.py]: fixed

File: sfs/mono/wfs.py (hunk 13 above)
Comment (fs446) [reply sfs/mono/wfs.py]: fixed

File: sfs/mono/wfs.py (hunk 11 above)
Comment (fs446) [reply sfs/mono/wfs.py]: fixed

File: sfs/mono/wfs.py (hunk 10 above)
Comment (fs446) [reply sfs/mono/wfs.py]: fixed

File: sfs/mono/wfs.py (hunk 7 above)
Comment (fs446) [reply sfs/mono/wfs.py]: fixed

File: sfs/mono/wfs.py (hunk 6 above)
Comment (fs446) [reply sfs/mono/wfs.py]: fixed

File: sfs/mono/wfs.py (hunk 8 above)
Comment (hagenw) [reply sfs/mono/wfs.py]: At the beginning I was also sceptical regarding the central approach and prefered ASCII math over proper LaTeX formula as ASCII was easier to read directly in the code. But in the end I realized that I was always using the online documentation to read the docstrings anyway.

I would also prefer to use something like `` :cite:`Spors2008` `` in the docstrings.

Comment (fs446) [reply sfs/mono/wfs.py]: > But if you read `` :cite:`Spors2008` `` in the source code, you can go to `doc/references.bib`, search for "Spors2008" and get all the information you need.

I'm not convinced. The code itself should contain a reference (DOI is enough, no fancy ASCII equation art) what actually is implemented with no need for a further file. This might change sometimes and nobody cares for the back link. Or the file might be not at hand, or or or...
I don't see a flaw to give just one more URL in the docstring additionally to the bibtex entry, which by its nature is fail safe, the other approach not!

Comment (fs446) [reply sfs/mono/wfs.py]: Thanks for the bib example, works charming! I've added the references.

File: sfs/mono/wfs.py (hunk 9 above)
Comment (fs446) [reply sfs/mono/wfs.py]: fixed

File: sfs/mono/wfs.py (hunk 12 above)
Comment (fs446) [reply sfs/mono/wfs.py]: fixed

File: sfs/mono/wfs.py (hunk 8 above)
Comment (mgeier) [reply sfs/mono/wfs.py]: It's true that in the case where you have previously installed the SFS Toolbox with `pip` and you don't have an internet connection, you will not be able to access the `references.bib` file (because it's not part of the wheel package).

But the DOI doesn't help either, because you don't have an internet connection, remember?

If you *do* have an internet connection, you should really just go to https://sfs-python.readthedocs.io/ and look at the documentation.

And if you know that you'll be offline at some point, you can get the PDF docs, which also include the references.
Or you can build the HTML docs locally.

File: sfs/mono/wfs.py [hunk 15]
```diff
@@ -155,8 +155,102 @@ def _point(omega, x0, n0, xs, c=None):
 
 
 def point_25d(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
+    r"""Driving function for 2.5-dimensional WFS of a virtual point source.
+
+    .. versionchanged:: 0.5
```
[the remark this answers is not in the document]
fs446: fixed

sfs/mono/wfs.py
```diff
@@ -155,8 +155,102 @@ def _point(omega, x0, n0, xs, c=None):
...
+    In versions <=0.4.0 we used the function that is now included in
```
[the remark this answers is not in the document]
fs446: fixed.

sfs/mono/wfs.py
```diff
@@ -155,8 +155,102 @@ def _point(omega, x0, n0, xs, c=None):
...
+        whether the corresponding secondary source is "active" or not.
+    secondary_source_function : callable
+        A function that can be used to create the sound field of a
+        single secondary source.  See `sfs.mono.synthesize()`.
+
+    Notes
+    -----
+    `point_25d()` derives 2.5D WFS from the 3D
+    Neumann-Rayleigh integral (i.e. the TU Delft approach), cf. e.g.
+    Eq. (3.10), (3.11) in :cite:`Start1997`.
```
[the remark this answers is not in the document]
fs446: fixed.

sfs/mono/wfs.py
```diff
@@ -155,8 +155,102 @@ def _point(omega, x0, n0, xs, c=None):
...
+
+
+point_3d = _point
+
+
+def point_25d_legacy(omega, x0, n0, xs, xref=[0, 0, 0], c=None, omalias=None):
     r"""Driving function for 2.5-dimensional WFS for a virtual point source.
 
+    .. versionadded:: 0.5
+    This driving function was used as `point_25d()` in versions <=0.4.0.
```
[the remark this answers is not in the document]
fs446: fixed.

sfs/mono/wfs.py
```diff
@@ -155,8 +155,102 @@ def _point(omega, x0, n0, xs, c=None):
...
+    We then changed to a `point_25d()` version requiring less assumptions on the
+    parameters.
```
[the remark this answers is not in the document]
fs446: good point, fixed.

fs446:
IMO it's not about having internet connection or not.

It's about the inconsistency providing information of the underlying
equation (either as ASCII or LaTex math) within the docstring as useful
raw data to be rendered somewhere else, but not doing the same principle
for the referencing part. The bibtex entry tag alone does not hold 
this information, I hope we agree on this.
The bibtex entry works fine and it is kind of failsafe within our environment, 
but if you would ask for following same principles, I'd expect something like
providing the DOI, or the bibtex code or plain ASCII within the docstring
and a rendering stage within sphinx. I could imagine that parsing a DOI
out of docstring and grabbing all information for creating a reference list
is something people ask for or even have done. Just an idea for future work.
The current framework is very good, but we should always try to improve.

sfs/mono/wfs.py
```diff
@@ -155,8 +155,101 @@ def _point(omega, x0, n0, xs, c=None):
...
+    -----
+    `point_25d()` derives 2.5D WFS from the 3D
+    Neumann-Rayleigh integral (i.e. the TU Delft approach), cf. e.g.
+    Eq. (3.10), (3.11) in :cite:`Start1997`.
+    Also cf. Eq. (2.137) in :cite:`Schultz2016`.
+
+    The theoretical link of `point_25d()` and `point_25d_legacy()` was
+    introduced as *unified WFS framework* in :cite:`Firtha2017`.
+
+    .. math::
```
mgeier: In the rendered docs, it kinda looks like this equation is mentioned here in reference to `Firtha2017`.

But it's really just the equation that's implemented in this function.

Same for the "legacy" version below.

Probably the order should be somehow changed?
Some additional text to introduce the equation?

mgeier:
> but not doing the same principle for the referencing part

Indeed.

We could, instead of using BibTeX, manually add plain Sphinx citations (https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#citations) to the docstrings.

Then your concern would be satisfied.

I just don't think it's worth it.

> The bibtex entry tag alone does not hold this information, I hope we agree on this.

We do.

> I could imagine that parsing a DOI out of docstring and grabbing all information for creating a reference list is something people ask for or even have done.

I don't know if anyone has asked for it, but I'm quite sure nobody has implemented it.

> The current framework is very good, but we should always try to improve.

Sure, there is room for improvement in the current situation.
But I think the BibTeX extension we are using is state-of-the-art, I didn't find any better extensions.

sfs/mono/wfs.py
fs446: good point, I've improved readability.

fs446:
> But I think the BibTeX extension we are using is state-of-the-art, I didn't find any better extensions.

Good job on that. I will remind myself from time to time to have a look on this issue. Can be closed for the moment.

fs446:
thx for your support!!

PR #125, opened 2019-03-12
new mono drivingfunction handling

fs446:
I'd make a suggestion to improve readability and code structure w.r.t. our driving function handling. 

Instead of current
`d, selection, secondary_source = sfs.mono.drivingfunction.wfs_25d_point(omega, array.x, array.n, xs)`
it changes to
`d, selection, secondary_source = sfs.mono.wfs.point_25d(omega, array.x, array.n, xs)`
and such like.

The introduced change splits the drivingfunction module into the specific SFS approaches, at the moment `wfs`, `sdm`, `esaedge`, `nfchoa`.
Meaningful for further progress could be the new modules `beamforming` and `panning` covering all aspects of farfield beams and pure gain / pure delay based SFS.

I'm still figuring what the best handling for `def secondary_source_point(omega, c):` would be. At the moment they appear in every SFS-module. This should be solved consistently with the time domain as well.


spors:
I would vote for the suggested changes!

sfs/mono/__init__.py
```diff
@@ -3,14 +3,23 @@
 .. autosummary::
     :toctree:
 
-    drivingfunction
+    esaedge
+    nfchoa
+    sdm
+    wfs
```
mgeier: This is alphabetic?

Shouldn't it be somehow "by importance"?

Or probably chronologically with respect to the SFS-Toolbox development?

sfs/mono/__init__.py
```diff
@@ -3,14 +3,23 @@
...
+
     source
```
mgeier: What about placing `source` before the driving functions?

sfs/mono/__init__.py
```diff
@@ -3,14 +3,23 @@
...
     soundfigure
```
mgeier: I'm not sure if it should happen within this PR, but what about moving this to the `wfs` module?

IOW, move `sfs.mono.soundfigure.wfs_3d_pw()` to `sfs.mono.wfs.soundfigure_something()`?

sfs/mono/esaedge.py
```diff
@@ -0,0 +1,400 @@
+"""Compute driving functions for various systems.
+
+.. include:: math-definitions.rst
+
+.. plot::
```
mgeier: None of this code is used in this file, it should be completely removed.

sfs/mono/nfchoa.py
```diff
@@ -0,0 +1,259 @@
...
+    :context: reset
+
+    import matplotlib.pyplot as plt
+    import numpy as np
+    import sfs
+
+    plt.rcParams['figure.figsize'] = 6, 6
+
+    xs = -1.5, 1.5, 0
+    xs_focused = -0.5, 0.5, 0
```
mgeier: `xs_focused` is unused.

sfs/mono/nfchoa.py
```diff
@@ -0,0 +1,259 @@
...
+    # normal vector for plane wave:
+    npw = sfs.util.direction_vector(np.radians(-45))
+    # normal vector for focused source:
+    ns_focused = sfs.util.direction_vector(np.radians(-45))
```
mgeier: `ns_focused` is unused.

sfs/mono/esaedge.py
```diff
@@ -0,0 +1,400 @@
...
+
+    def plot(d, selection, secondary_source):
+        p = sfs.mono.synthesize(d, selection, array, secondary_source, grid=grid)
+        sfs.plot.soundfield(p, grid)
+        sfs.plot.loudspeaker_2d(array.x, array.n, selection * array.a, size=0.15)
+
+"""
+
+import numpy as np
+from numpy.core.umath_tests import inner1d  # element-wise inner product
```
mgeier: Unused `import`.

sfs/mono/esaedge.py
```diff
@@ -0,0 +1,400 @@
...
+from scipy.special import jn, hankel2
+from .. import util
+from .. import default
```
mgeier: Unused `import`.

sfs/mono/nfchoa.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def plot(d, selection, secondary_source):
+        p = sfs.mono.synthesize(d, selection, array, secondary_source, grid=grid)
+        sfs.plot.soundfield(p, grid)
+        sfs.plot.loudspeaker_2d(array.x, array.n, selection * array.a, size=0.15)
+
+"""
+
+import numpy as np
+from numpy.core.umath_tests import inner1d  # element-wise inner product
```
mgeier: Unused `import`.

sfs/mono/nfchoa.py
```diff
@@ -0,0 +1,259 @@
...
+from scipy.special import jn, hankel2
```
mgeier: `jn` is unused.

sfs/mono/nfchoa.py
```diff
@@ -0,0 +1,259 @@
...
+from .. import util
+from .. import default
```
mgeier: Unused `import`.

sfs/mono/sdm.py
```diff
@@ -0,0 +1,284 @@
...
+    :context: reset
+
+    import matplotlib.pyplot as plt
+    import numpy as np
+    import sfs
+
+    plt.rcParams['figure.figsize'] = 6, 6
+
+    xs = -1.5, 1.5, 0
+    xs_focused = -0.5, 0.5, 0
```
mgeier: Unused.

sfs/mono/sdm.py
```diff
@@ -0,0 +1,284 @@
...
+    # normal vector for plane wave:
+    npw = sfs.util.direction_vector(np.radians(-45))
+    # normal vector for focused source:
+    ns_focused = sfs.util.direction_vector(np.radians(-45))
```
mgeier: Unused.

sfs/mono/sdm.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def plot(d, selection, secondary_source):
+        p = sfs.mono.synthesize(d, selection, array, secondary_source, grid=grid)
+        sfs.plot.soundfield(p, grid)
+        sfs.plot.loudspeaker_2d(array.x, array.n, selection * array.a, size=0.15)
+
+"""
+
+import numpy as np
+from numpy.core.umath_tests import inner1d  # element-wise inner product
```
mgeier: Unused.

sfs/mono/sdm.py
```diff
@@ -0,0 +1,284 @@
...
+from scipy.special import jn, hankel2
```
mgeier: `jn` is unused.

sfs/mono/sdm.py
```diff
@@ -0,0 +1,284 @@
...
+from .. import util
+from .. import default
```
mgeier: Unused.

sfs/mono/wfs.py
```diff
@@ -0,0 +1,582 @@
...
+    def plot(d, selection, secondary_source):
+        p = sfs.mono.synthesize(d, selection, array, secondary_source, grid=grid)
+        sfs.plot.soundfield(p, grid)
+        sfs.plot.loudspeaker_2d(array.x, array.n, selection * array.a, size=0.15)
+
+"""
+
+import numpy as np
+from numpy.core.umath_tests import inner1d  # element-wise inner product
+from scipy.special import jn, hankel2
```
mgeier: `jn` is unused.

sfs/mono/wfs.py
```diff
@@ -0,0 +1,582 @@
...
+from .. import util
+from .. import default
```
mgeier: Unused.

mgeier:
Looks like you didn't delete `sfs/mono/drivingfunction.py`!

sfs/mono/__init__.py
hagenw: Alphabetic is the easiest to understand and you don't have to fight about which is more important.

hagenw:
> Thanks, this looks really nice!
> 
> I would put the secondary source functions directly into the `sfs.mono` namespace, i.e. define them in `sfs/mono/__init__.py`, right next to `sfs.mono.synthesize()`.

I wasn't aware of this before. I thought `__init__.py` is only for imports or in some cases even completely empty. I wouldn't have looked for any function definitions there.

Do you put all stuff there that doesn't fit into another file?

sfs/mono/sdm.py
```diff
@@ -0,0 +1,284 @@
...
+    Eq.(3.79) from :cite:`Ahrens2012`.
+
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        array = sfs.array.linear(32, 0.2, orientation=[0, -1, 0])
+        d, selection, secondary_source = (
+            sfs.mono.sdm._25d(
```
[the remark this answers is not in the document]
fs446: weird, thought I've completely checked the sphinx output, very obviously not, thx.

sfs/mono/sdm.py
```diff
@@ -0,0 +1,284 @@
...
+                omega, array.x, array.n, npw, [0, -1, 0]))
```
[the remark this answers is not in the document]
fs446: fixed to suggestion

sfs/mono/__init__.py
fs446: alphabetic was indeed intentional. I don't have a preference.

sfs/mono/__init__.py
fs446: yes, that seems to be a good categorizing style.

sfs/mono/__init__.py
fs446: moved to sfs.mono.wfs.soundfigure_2d()
I am not sure, it might even be a _3d scenario, can you @spors validate naming.

Note: this code is not tested at all in the moment, so I'm going to open a new issue for this. But for now the handling would be more consistent.

mgeier:
@hagenw 

> I wasn't aware of this before. I thought `__init__.py` is only for imports or in some cases even completely empty. I wouldn't have looked for any function definitions there.

Well all non-module (and non-package) contents of a Python "package" (i.e. defined by a directory) has to come from `__init__.py`. Whether it's imported from somewhere else or defined right there doesn't really make a difference.

> Do you put all stuff there that doesn't fit into another file?

I think that's a matter of taste. Some people define their stuff right there, but other make separate submodules (often "hidden", like `_stuff.py`) and then import everything with

```python
from ._stuff import *
```

Neither of those methods are really nice, but that's how it is in Python ...

I slightly prefer putting definitions directly into `__init__.py`, if only to avoid having to come up with a name for a "hidden" submodule.

mgeier:
Just one random example for an `__init__.py` from CPython itself: https://github.com/python/cpython/blob/master/Lib/collections/__init__.py

hagenw:
OK, wasn't completely aware of the `__init__` stuff.
And you are right, none of the solutions are super nice.

fs446:
With the new handling we could adapt the init source code within the docstring examples to the specific SFS approach, such as SDM gets a linear SSD and so on.

mgeier:
@fs446 Yeah that's a good idea, but probably for a separate, later PR?

fs446:
> OK, wasn't completely aware of the `__init__` stuff.
> And you are right, none of the solutions are super nice.

I'm not a big friend of it either, but I'm sure we find an elegant place for it when the changes evolve. At the moment the handling for most user friendly API is good, I like the API very much now.

We furthermore should think about the SFS categories as well:
As I discussed today with @narahahn basically all Mode Matching approaches, such as SDM/NFC-HOA are equivalent with the Equivalent Scattering Approach for the specific SSD type. So in fact we currently dealing with three ESA types with weird names ;-) (SDM, NFC-HOA, Edge) and the WFS type as ray-based/HF/Far Approx of the explicit solutions rather independent of a SSD geometry.
For the moment I'd vote to go with the currently proposed name scheme, but as soon as we implement other panning, beamforming stuff we should seriously think about this categorizing.

narahahn:
First of all, thank's a lot for the changes. I really like them.

> As I discussed today with @narahahn basically all Mode Matching approaches, such as SDM/NFC-HOA are equivalent with the Equivalent Scattering Approach for the specific SSD type. So in fact we currently dealing with three ESA types with weird names ;-) (SDM, NFC-HOA, Edge) and the WFS type as ray-based/HF/Far Approx of the explicit solutions rather independent of a SSD geometry.
> For the moment I'd vote to go with the currently proposed name scheme, but as soon as we implement other panning, beamforming stuff we should seriously think about this categorizing.

For now, I'd also keep the current naming scheme.
Despite the similarity among the mode-matching approaches,
the implementations have been evolved and optimized in different ways
mostly due to the specific SSDs under consideration.
IMHO, each deserves its own submodule.

I have one suggestion, though.
How about change `esaedge` to `esa` and
put the driving function for edge SSD under the module?  Like `esa.point_edge()`.
Someday, we might want to add other `esa` driving functions for different types of SSD,
that are not covered by `sdm` and `nfchoa`.

sfs/mono/sdm.py
```diff
@@ -0,0 +1,263 @@
...
+    The secondary sources have to be located on the x-axis (y0=0).
+    Driving function from :cite:`Spors2010`, Eq.(24).
+
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        array = sfs.array.linear(32, 0.2, orientation=[0, -1, 0])
+        d, selection, secondary_source = (
```
mgeier:
```python
        d, selection, secondary_source = sfs.mono.sdm.point_25d(
            omega, array.x, array.n, xs, [0, -1, 0])
```

mgeier:
In `sfs/mono/__init__.py` I would move each of the `import` statements either to the top or the bottom (as mentioned in individual comments).

Then you should be free to choose any ordering for the functions.

I would not define `secondary_source_point()` or `secondary_source_line()` first, because they are the most boring functions in this module.

sfs/mono/esaedge.py
fs446: da37636be625e9dfbb5fa98ceee3b2e336037006

sfs/mono/nfchoa.py
fs446: 23c4f8e575f3d7a63283412ce7de1e43d55d347d

sfs/mono/nfchoa.py
fs446: 23c4f8e575f3d7a63283412ce7de1e43d55d347d

sfs/mono/esaedge.py
fs446: 3779e598b626bb148d4c267c7bc34973583ce772

sfs/mono/esaedge.py
fs446: 3779e598b626bb148d4c267c7bc34973583ce772

sfs/mono/nfchoa.py
fs446: 0615628109c14ffd53dce6506b1daa9af6e59e71

sfs/mono/nfchoa.py
fs446: 0615628109c14ffd53dce6506b1daa9af6e59e71

sfs/mono/nfchoa.py
fs446: 0615628109c14ffd53dce6506b1daa9af6e59e71

sfs/mono/sdm.py
fs446: 0c86e4d1bfc1a50c12e8f1f0ab18b7715121ac70

sfs/mono/sdm.py
fs446: 0c86e4d1bfc1a50c12e8f1f0ab18b7715121ac70

sfs/mono/sdm.py
fs446: aed7ab787c18b904e3d39f08a1b5f8725ee68596

fs446:
Idea from @narahahn  
> I have one suggestion, though.
> How about change `esaedge` to `esa` and
> put the driving function for edge SSD under the module? Like `esa.point_edge()`.
> Someday, we might want to add other `esa` driving functions for different types of SSD,
> that are not covered by `sdm` and `nfchoa`.

is great. I've changed esaedge to esa in 7424338f93fbd01cdf1aa6f23f97895b06c6cada
renamed the functions, which is even more consistent now and left some
docstring comments for better understanding what is/should going on in this module.

PR #126, opened 2019-03-12
new time driving function handling

fs446:
Same idea as of #125

Additionally compared to frequency domain, here the handling of `def driving_signals` and `def apply_delays` should be handled more elegant. I'm currently testing some ideas.


mgeier:
Very nice!

I don't see a problem with the different variants of `driving_signal()`.

You could move `apply_delays()` to `sfs.time.apply_delays()`, because it is more general than WFS.

I didn't make a detailed review yet, most of the things from #125 apply here, too.

sfs/time/__init__.py
```diff
@@ -3,15 +3,16 @@
 .. autosummary::
     :toctree:
 
-    drivingfunction
+    wfs
+    nfchoa
+
     source
```
mgeier: This should be the first entry, as in #125.

mgeier:
@narahahn Sorry, I've already merged this PR.

For the return value documentation, you seem to have already a commit for that: 9155049657a4e805d929cfd4cc517d044a9f8d6e. You just have to rebase it to `master`.

Regarding the other comments, @fs446 could you please address them in a separate PR?

mgeier:
@narahahn Those other comments seem outdated anyway, right?

narahahn:
> @narahahn Sorry, I've already merged this PR.

No problem.

> For the return value documentation, you seem to have already a commit for that: [9155049](https://github.com/sfstoolbox/sfs-python/commit/9155049657a4e805d929cfd4cc517d044a9f8d6e). You just have to rebase it to `master`.

I just opened a PR.

> @narahahn Those other comments seem outdated anyway, right?

Yes, they do.

PR #130, opened 2019-03-13
Allow "import sfs.time as sfs"

mgeier:
I'm not sure if this is helpful or confusing, could be both ...

The idea is that *most of the time*, we want to use *either* time domain *or* frequency domain. There are much fewer cases where we want to use both in the same script/notebook (the "mirror image" notebook is an example for that).

To make it easier for the many cases, we could change the import a bit:

```python
import sfs.time as sfs
```

... and then we don't ever need to use `time` in the rest of the script/notebook anymore.

What do you think about this?
Is this silly?

### DIFF
--- a/doc/examples/animations_pulsating_sphere.py
+++ b/doc/examples/animations_pulsating_sphere.py
@@ -1,5 +1,5 @@
 """Animations of pulsating sphere."""
-import sfs
+import sfs.mono as sfs
 import numpy as np
 from matplotlib import pyplot as plt
 from matplotlib import animation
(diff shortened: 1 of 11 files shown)


fs446:
> What do you think about this?
> Is this silly?

Just double checked, works as expected. I'd vote to merge this feature into master. We get an even cleaner API for the discussed cases, which looks nice. I don't see any flaws at the moment.

mgeier:
I think this is not really necessary since we chose the very short module names `fd`/`td` in #135.

We can re-evaluate it at a later time if we feel the need for it.

PR #133, opened 2019-03-13
Use \mathop instead of \operatorname

mgeier:
TBH, I don't understand what the difference is between the two, but if I use `\operatorname`, I get an "ffi ligature" instead of a "lowercase greek letter delta" in the PDF.

### DIFF
--- a/doc/math-definitions.rst
+++ b/doc/math-definitions.rst
@@ -5,7 +5,7 @@
 .. rst-class:: hidden
 .. math::
 
-    \gdef\dirac#1{\operatorname{\delta}\left(#1\right)}
+    \gdef\dirac#1{\mathop{{}\delta}\left(#1\right)}
     \gdef\e#1{\operatorname{e}^{#1}}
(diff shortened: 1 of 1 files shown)


hagenw:
As I understood https://tex.stackexchange.com/a/84308 `\operatorname` is the safer choice. Have you checked if the spacing etc. is identical in the compiled PDF?
I do not understand what you mean by "ffi ligature" instead of "lowercase greek letter delta".

mgeier:
for example:

image

hagenw:
This is strange.
The same definition of `\dirac` works for https://github.com/sfstoolbox/theory without any problems, see
dirac
The definition is:
```latex
\def \dirac          #1{\operatorname{\delta}\left(#1\right)}
```
So, the only difference is that we use `\gdef` in sfs-python instead.

mgeier:
No, it's not the only difference.

We are also using a different math font (with the `mathpazo` package), which causes the problem to appear.

hagenw:
Ah, ok.
I can confirm that it works with your changes. So it's fine for to use `\mathop` then.

PR #135, opened 2019-03-14
Change mono/time to fd/td

mgeier:
Just to get an idea how this would look like.

The related discussion is in #25.

The "non-fancy" part in the branch name means that this is *not* based on #130.


hagenw:
I guess you have to add some justification, why you prefer the non obvious abbreviations `fd` and `td` over `mono` and `freq` as they were not selected in #25.

mgeier:
@hagenw Please see https://github.com/sfstoolbox/sfs-python/issues/25#issuecomment-472806617.

PR #136, opened 2019-03-14
Change mono/time to fd/td, an alternative view

mgeier:
This is similar to #135, but based on #130.

Again, just to get an idea how this would look like.


mgeier:
Superseded by #135.

Issue #107, opened 2019-03-07
mirror image notebook: RuntimeWarning: divide by zero 


mgeier:
Currently, there are several ugly warnings in this notebook: http://python.sfstoolbox.org/en/latest/examples/mirror-image-source-model.html

That's the first bunch:

```
/home/docs/checkouts/readthedocs.org/user_builds/sfs-python/conda/latest/lib/python3.7/site-packages/sfs/mono/source.py:85: RuntimeWarning: divide by zero encountered in true_divide
  return 1 / (4*np.pi) * np.exp(-1j * k * r) / r
/home/docs/checkouts/readthedocs.org/user_builds/sfs-python/conda/latest/lib/python3.7/site-packages/sfs/mono/source.py:85: RuntimeWarning: invalid value encountered in true_divide
  return 1 / (4*np.pi) * np.exp(-1j * k * r) / r
/home/docs/checkouts/readthedocs.org/user_builds/sfs-python/conda/latest/lib/python3.7/site-packages/sfs/mono/source.py:398: RuntimeWarning: invalid value encountered in multiply
  p += strength * point(omega, position, grid, c)
```

And two more further down:

```
/home/docs/checkouts/readthedocs.org/user_builds/sfs-python/conda/latest/lib/python3.7/site-packages/sfs/time/source.py:82: RuntimeWarning: divide by zero encountered in true_divide
  weights = 1 / (4 * np.pi * r)
/home/docs/checkouts/readthedocs.org/user_builds/sfs-python/conda/latest/lib/python3.7/site-packages/sfs/time/source.py:87: RuntimeWarning: invalid value encountered in multiply
  data, left=0, right=0)
```

It would be nice to get rid of those warnings.

If the division by zero is justified, we should silence the warning.

One example for ignoring a warning:

https://github.com/sfstoolbox/sfs-python/blob/d4ac3103c2addc8d2a253fd42cf4cbcc4121f238/sfs/plot.py#L388-L389

Another one:

https://github.com/sfstoolbox/sfs-python/blob/d4ac3103c2addc8d2a253fd42cf4cbcc4121f238/sfs/util.py#L366-L367

mgeier:
This was resolved in #139.

PR #139, opened 2019-03-14
Hide warnings for a few undefined points

hagenw:
This suppresses the warnings mentioned in #107.

But I think the current solution is very ugly and covers only a few of the places where this can occur.
The problem is that a lot of our sources, e.g. a point source have some points where they are not defined and it is valid to have `NaN`, `Inf`, divide by zero or other stuff. But it will be a huge effort to find every line where this has to be fixed.

Another solution would be to generally allow for division by zero or multiplication by `NaN` in the whole toolbox, but this is also not nice.

Has somebody an idea how to better handle undefined points in numpy?

### DIFF
--- a/sfs/fd/source.py
+++ b/sfs/fd/source.py
@@ -84,7 +84,7 @@ def point(omega, x0, grid, *, c=None):
     r = np.linalg.norm(grid - x0)
     # If r is 0, the sound pressure is complex infinity
     numerator = np.exp(-1j * k * r) / (4*np.pi)
-    with np.errstate(divide='ignore'):
+    with np.errstate(invalid='ignore', divide='ignore'):
         return numerator / r
 
 
--- a/sfs/fd/source.py
+++ b/sfs/fd/source.py
@@ -82,7 +82,9 @@ def point(omega, x0, grid, *, c=None):
     grid = util.as_xyz_components(grid)
 
     r = np.linalg.norm(grid - x0)
-    return 1 / (4*np.pi) * np.exp(-1j * k * r) / r
+    # `r` can get `0`, which is fine in this case
+    with np.errstate(invalid='ignore', divide='ignore'):
+        return 1 / (4*np.pi) * np.exp(-1j * k * r) / r
 
 
 def point_velocity(omega, x0, grid, *, c=None, rho0=None):
@@ -395,7 +397,9 @@ def point_image_sources(omega, x0, grid, L, *, max_order, coeffs=None, c=None):
     p = 0
     for position, strength in zip(xs, source_strengths):
         if strength != 0:
-            p += strength * point(omega, position, grid, c=c)
+            # `point` can return NaN, which is fine
+            with np.errstate(invalid='ignore'):
+                p += strength * point(omega, position, grid, c=c)
 
     return p
 
--- a/sfs/fd/source.py
+++ b/sfs/fd/source.py
@@ -82,9 +82,10 @@ def point(omega, x0, grid, *, c=None):
     grid = util.as_xyz_components(grid)
 
     r = np.linalg.norm(grid - x0)
-    # `r` can get `0`, which is fine in this case
-    with np.errstate(invalid='ignore', divide='ignore'):
-        return 1 / (4*np.pi) * np.exp(-1j * k * r) / r
+    # If r is 0, the sound pressure is complex infinity
+    numerator = np.exp(-1j * k * r) / (4*np.pi)
+    with np.errstate(divide='ignore'):
+        return numerator / r
 
 
 def point_velocity(omega, x0, grid, *, c=None, rho0=None):
@@ -397,7 +398,7 @@ def point_image_sources(omega, x0, grid, L, *, max_order, coeffs=None, c=None):
     p = 0
     for position, strength in zip(xs, source_strengths):
         if strength != 0:
-            # `point` can return NaN, which is fine
+            # point can be complex infinity
             with np.errstate(invalid='ignore'):
                 p += strength * point(omega, position, grid, c=c)
 
--- a/sfs/td/source.py
+++ b/sfs/td/source.py
@@ -79,16 +79,17 @@ def point(xs, signal, observation_time, grid, c=None):
     if c is None:
         c = default.c
     r = np.linalg.norm(grid - xs)
-    # evaluate g over grid (`r` can get `0`, which is fine)
+    # If r is +-0, the sound pressure is +-infinity
     with np.errstate(divide='ignore'):
         weights = 1 / (4 * np.pi * r)
     delays = r / c
     base_time = observation_time - signal_offset
-    # Result can be NaN, which is fine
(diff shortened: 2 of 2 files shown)


mgeier:
> Has somebody an idea how to better handle undefined points in numpy?

No, I think allowing NaN and +-Inf in those cases makes sense.
Another possibility would be to use "masked arrays", but I don't think those are worth the added complexity.

> But I think the current solution is very ugly and covers only a few of the places where this can occur.

I don't think that's a problem as long as the resulting NaN and +-Inf values are theoretically meaningful.
There is no need to cover all cases now, we can just keep adding those "ignore" things as we discover new warnings.

> generally allow for division by zero or multiplication by NaN in the whole toolbox

We shouldn't do that.
There are cases where these warnings show an actual bug or mistake.
We should only silence the warnings for situations where we know that the resulting values make sense based on the theory.

sfs/mono/source.py
```diff
@@ -82,7 +82,9 @@ def point(omega, x0, grid, c=None):
     grid = util.as_xyz_components(grid)
 
     r = np.linalg.norm(grid - x0)
-    return 1 / (4*np.pi) * np.exp(-1j * k * r) / r
+    # `r` can get `0`, which is fine in this case
```
mgeier: I don't really like the backticks here, because they are not used for actual formatting.

"which is fine in this case" sounds a bit strange to me. Who says it's fine?

The fact that "we allow it" or "it's fine" is already explicit because we use this "ignore" thing, isn't it?

What about:

```python
# If r is 0, the sound pressure is complex infinity
```

or something like that?

This explains the problem, the solution is already visible and needs no further comment.

We could (should?) probably also mention this in the docstring, right?

sfs/mono/source.py
```diff
@@ -395,7 +397,9 @@ def point_image_sources(omega, x0, grid, L, max_order, coeffs=None, c=None):
     p = 0
     for position, strength in zip(xs, source_strengths):
         if strength != 0:
-            p += strength * point(omega, position, grid, c)
+            # `point` can return NaN, which is fine
```
mgeier: Is it really NaN?
Isn't it "complex infinity"?

sfs/time/source.py
```diff
@@ -79,13 +79,16 @@ def point(xs, signal, observation_time, grid, c=None):
     if c is None:
         c = default.c
     r = np.linalg.norm(grid - xs)
-    # evaluate g over grid
-    weights = 1 / (4 * np.pi * r)
+    # evaluate g over grid (`r` can get `0`, which is fine)
```
mgeier: What about

```python
# If r is +- 0, the sound pressure is +- infinity
```

BTW, if the comment "evaluate g over grid" is really necessary, it should go directly before the calculation, not before the `with` statement.

sfs/time/source.py
```diff
@@ -79,13 +79,16 @@ def point(xs, signal, observation_time, grid, c=None):
...
+    with np.errstate(divide='ignore'):
+        weights = 1 / (4 * np.pi * r)
     delays = r / c
     base_time = observation_time - signal_offset
-    return weights * np.interp(base_time - delays,
-                               np.arange(len(data)) / samplerate,
-                               data, left=0, right=0)
+    # Result can be NaN, which is fine
```
mgeier: What's the actual problem here?
If `weights` is +- inf?

In that case you should probably move the `np.interp()` call out of the `with` statement.

sfs/mono/source.py
hagenw: Is complex infinity a defined term?
When I calculate the pressure of a point source at its source position, I'm getting a mixture of `nan` and `inf`:
```python
>>> sfs.mono.source.point(omega, x0, x0)
(inf+nanj)
```

sfs/time/source.py
hagenw: `weights` are the output of a possible `1/0` division, which means it can contain `(inf+nanj)`.

sfs/mono/source.py
mgeier: I'm not an expert, but I think that's NumPy's definition of "complex infinity". I couldn't find any mention of this in the NumPy docs, though.

It is a "defined term", but it's surprisingly hard to find information about it. Here's one example: http://oeis.org/wiki/Division_by_zero

FWIW, the Wolfram language has this as a separate symbol: https://reference.wolfram.com/language/ref/ComplexInfinity.html

sfs/time/source.py
mgeier: OK, then the `np.interp()` call should not be in the `with` statement.

sfs/mono/source.py
```diff
@@ -82,7 +82,9 @@ def point(omega, x0, grid, c=None):
     grid = util.as_xyz_components(grid)
 
     r = np.linalg.norm(grid - x0)
-    return 1 / (4*np.pi) * np.exp(-1j * k * r) / r
+    # `r` can get `0`, which is fine in this case
+    with np.errstate(invalid='ignore', divide='ignore'):
+        return 1 / (4*np.pi) * np.exp(-1j * k * r) / r
```
mgeier: We could probably also be more obvious here and put the numerator in a temporary variable (before opening the `with` statement).

The we could have only the critical division within the `with` statement.

```python
numerator = np.exp(-1j * k * r) / (4*np.pi)
with np.errstate(invalid='ignore', divide='ignore'):
    return numerator / r
```

Would this make sense?

mgeier: If we do it like this, we probably don't even need a separate comment explaining it?

hagenw: Good idea, then the `invalid='ignore'` should also be not needed.

sfs/mono/source.py
hagenw: Thanks, for looking it up. Changed it to your proposal.

sfs/mono/source.py
hagenw: Changed it to
```python
# point can be complex infinity
```

sfs/time/source.py
hagenw: Changed it accordingly, and removed "evaluate g over grid"

sfs/time/source.py
hagenw: Changed it to
```python
    points_at_time = np.interp(base_time - delays,
                               np.arange(len(data)) / samplerate,
                               data, left=0, right=0)
    # weights can be +-infinity
    with np.errstate(invalid='ignore'):
        return weights * points_at_time
```

mgeier:
Thanks for the changes!

It looks like `invalid='ignore'` is still needed:

```pycon
>>> import sfs
>>> sfs.fd.source.point(100, [6, 6, 6], [6, 6, 6])
[...]/sfs/fd/source.py:88: RuntimeWarning: invalid value encountered in cdouble_scalars
  return numerator / r
(inf+nanj)
>>>

hagenw:
Mh, strange for the notebook example I got no warning.
I will add it back.

PR #140, opened 2019-03-14
Keyword-only arguments everywhere

mgeier:
Keyword-only arguments are a nice Python 3 feature to force the user to use keyword arguments, avoiding very long sequences of positional parameters.

The good thing is that if it gets annoying, the requirements can be weakened without causing backwards-compatibility issues.

Adding them is typically backwards-incompatible.

### DIFF
--- a/sfs/array.py
+++ b/sfs/array.py
@@ -84,7 +84,7 @@ def as_secondary_source_distribution(arg, **kwargs):
     return SecondarySourceDistribution(x, n, a)
 
 
-def linear(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
+def linear(N, spacing, *, center=[0, 0, 0], orientation=[1, 0, 0]):
     """Return linear, equidistantly sampled secondary source distribution.
 
     Parameters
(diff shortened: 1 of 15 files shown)


sfs/array.py
```diff
@@ -84,7 +84,7 @@ def as_secondary_source_distribution(arg, **kwargs):
     return SecondarySourceDistribution(x, n, a)
 
 
-def linear(N, spacing, center=[0, 0, 0], orientation=[1, 0, 0]):
+def linear(N, spacing, *, center=[0, 0, 0], orientation=[1, 0, 0]):
```
hagenw: What does the `*` do here?

mgeier: That's the notation for keyword-only arguments, see https://www.python.org/dev/peps/pep-3102/.

hagenw: OK, I see, but why do we need this already, even `linear` has no variable arguments?
If we introduce them later, we could still add `*`, couldn't we?

hagenw: Do I get it right and `*` forces the keyword arguments to be used only as keywords?
So you have to write `def linear(N, spacing, center=[0, 1, 0])` and you are not allowed to write `def linear(N, spacing, [0, 1, 0])`.
Then I get the point.

mgeier: > forces the keyword arguments to be used only as keywords

Exactly. That's what "keyword-only" means.

> you are not allowed to write `linear(N, spacing, [0, 1, 0])`

Exactly.

And if we introduce them later, it will be a breaking change.
If we decide to remove them later, it's no problem, because removing them is backwards-compatible.

hagenw: :+1:

Issue #26, opened 2016-12-11
Rename plotting functions


mgeier:
Since we can plot different types of soundfields (pressure, velocity), the name `sfs.plot.soundfield()` is ambiguous.

There is a function called `sfs.plot.level()`, which means we could probably use something like `sfs.plot.amplitude()` instead.
Or probably it should be something with "pressure" in its name?

The function `sfs.plot.loudspeaker_3d()` doesn't actually plot loudspeakers.

Also, in most cases multiple loudspeakers will be plotted, so probably using the plural would be appropriate?

Plotting loudspeakers in 2D will probably be the default, therefore the name `loudspeaker_2d()` seems clumsy and should probably be renamed to simply `loudspeaker()` (or the plural of it).

Another possibility would be that we make two separate submodules for 2D and 3D plotting, what about that?

spors:
The function  `sfs.plot.soundfield()` plots the real value of a complex field. It is hence useful in the context of frequency-domain simulations. But I also used it for other fields (e.g. velocity). Perhaps I would be a good idea to have low-level plotting functions for scalar (and vector) fields and specialized functions for pressure etc.

In fact `sfs.plot.level()` plots the logarithmic magnitude of a (complex) field. For a logarithmic plot (in dB) the term level is correct. But we should allow to supply a nomalization value (e.g. p_0 = 2e-5 for SPL). If plotted linear we should name it  `sfs.plot.magnitude()`.

I would vote for 2D/3D submodules. 3D plotting will always be a bit experimental.

fs446:
I'd vote for:
with 2D/3D submodules
sfs.plot.amplitude() -> plots the real part of a complex scalar / vector field
sfs.plot.magnitude() -> plots the magnitude (i.e. abs) of a complex scalar / vector field
sfs.plot.level() -> plots the level (i.e. in dB) of a complex scalar / vector field

some further thoughts on it:
-shortened versions could be 
sfs.plot.lin()
sfs.plot.abs()
sfs.plot.db()
-level() / db() should accept a reference value (e.g. p_0) as mentioned above
-'physical unit'-string should be accepted for proper labeling
-amplitude() / lin() could accept a flag if plotting either the real or the imag part (for whatever reason this might be useful, since most conventions use real)
-should the functions accept both, scalar and vector fields and do an appropriate plotting job by checking array dimension??
-if no: a meaningful handling could be sfs.plot.scalar_lin() vs. sfs.plot.vector_lin() and so on
[...]

PR #145, opened 2019-03-15
Split "plot" module into "plot2d" and "plot3d"

mgeier:
This implements some, but by far not all ideas from #26.


fs446:
Charming! Works as expected. Go for it.

PR #156, opened 2019-11-21
add point source with far-field directivity of circular piston

fietew:
Adds new source type with far-field directivity  of circular piston radiator

```Python
import sfs
import numpy as np
plt.rcParams['figure.figsize'] = 8, 4.5  # inch

x0 = 1.5, 1, 0
n0 = 0, -1, 0
R = 1.0  # meters
f = 500  # Hz
omega = 2 * np.pi * f

normalization_point = 4 * np.pi

grid = sfs.util.xyz_grid([-2, 3], [-1, 2], 0, spacing=0.02)

p = sfs.fd.source.point_circular_piston(omega, x0, n0, grid, R)
sfs.plot2d.amplitude(p * normalization_point, grid)
plt.title("Circular Piston Radiator at {} m".format(x0))
```

### DIFF
--- a/sfs/fd/source.py
+++ b/sfs/fd/source.py
@@ -224,6 +224,68 @@ def point_dipole(omega, x0, n0, grid, *, c=None):
     return 1 / (4 * _np.pi) * (1j * k + 1 / r) * _np.inner(offset, n0) / \
         _np.power(r, 2) * _np.exp(-1j * k * r)
 
+def point_circular_piston(omega, x0, n0, grid, R, *, c=None):
+    r"""Point source with circular piston far-field directivity.
+
+    Parameters
+    ----------
+    omega : float
+        Frequency of source.
+    x0 : (3,) array_like
+        Position of source.
+    n0 : (3,) array_like
+        Normal vector (direction) of the circular piston.
+    grid : triple of array_like
+        The grid that is used for the sound field calculations.
+        See `sfs.util.xyz_grid()`.
+    R : float
+        Radius of circular piston.
+    c : float, optional
+        Speed of sound.
+    Returns
+    -------
+    numpy.ndarray
+        Sound pressure at positions given by *grid*.
+
+    Notes
+    -----
+    .. math::
+
+        G(\x-\x_0,\w) = \frac{1}{2\pi} 
+            \frac{J_1(\wc R \sin(\Theta))}{\wc R \sin(\Theta)}
+            \frac{\e{-\i\wc|\x-\x_0|}}{|\x-\x_0|}
+    with 
+
+    .. math::
+
+        \cos(\Theta) =
+            \frac{\langle \mathbf{x} - \mathbf{x}_0 | \mathbf{n}_0 \rangle}
+                {|\mathbf{x} - \mathbf{x}_0|}
+
+    Examples
+    --------
+    .. plot::
+        :context: close-figs
+
+        n0 = 0, -1, 0
+        R = 1.0
+        p = sfs.fd.source.point_circular_piston(omega, x0, n0, grid, R)
+        sfs.plot2d.amplitude(p * normalization_point, grid)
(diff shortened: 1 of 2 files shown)


mgeier:
The machine is complaining:

```
$ python3 -m pycodestyle sfs/fd/source.py
sfs/fd/source.py:227:1: E302 expected 2 blank lines, found 1
sfs/fd/source.py:254:39: W291 trailing whitespace
sfs/fd/source.py:257:9: W291 trailing whitespace
$ python3 -m pycodestyle sfs/util.py
sfs/util.py:580:1: E302 expected 2 blank lines, found 1
```

And https://travis-ci.org/sfstoolbox/sfs-python/jobs/615577318 contains:

    /home/travis/build/sfstoolbox/sfs-python/sfs/fd/source.py:docstring of sfs.fd.source.point_circular_piston:20: WARNING: Explicit markup ends without a blank line; unexpected unindent.

Note that the line number is meant within the docstring.
I guess the problem is a missing blank line before `Returns`.

sfs/fd/source.py
```diff
@@ -224,6 +224,69 @@ def point_dipole(omega, x0, n0, grid, *, c=None):
...
+    .. math::
+
+        G(\x-\x_0,\w) = \frac{1}{2\pi} 
+            \frac{J_1(\wc R \sin(\Theta))}{\wc R \sin(\Theta)}
+            \frac{\e{-\i\wc|\x-\x_0|}}{|\x-\x_0|}
+    with 
+
+    .. math::
+
+        \cos(\Theta) =
```
mgeier: Sorry if this is a naive question, but why are you mentioning `cos()` here when `sin()` is used in the equation above?

fietew: It's because, the cosine can be expressed by the scalar product of the involved vectors, while the sine has to be expressed ``sqrt(1 - cos^2())``, which is also not fully correct because it could also be ``-sqrt(1 - cos^2())``.

mgeier: I see. Probably you could add those explanations to the docs then, to avoid people thinking this is a mistake?

sfs/util.py
```diff
@@ -555,6 +556,28 @@ def spherical_hn2(n, z):
     return spherical_jn(n, z) - 1j * spherical_yn(n, z)
 
 
+def jinc(x):
+    r"""Circular counterpart of Sinc function a.k.a. Jinc function
```
mgeier: The third dot messes up the short description used in the summary table.

Also:

```
$ python3 -m pydocstyle sfs/util.py
sfs/util.py:559 in public function `jinc`:
        D400: First line should end with a period (not 'n')
```

sfs/util.py
```diff
@@ -555,6 +556,28 @@ def spherical_hn2(n, z):
...
+
+    Parameters
+    ----------
+    x : array_like
+        Argument of the Jinc function.
+
+    """
+
+    x = np.asanyarray(x)
+    y = where(x == 0, 1.0e-20, x)
```
mgeier: I can't tell if this is the right thing to do.

Is there a reference implementation or a literature reference you could mention?

sfs/fd/source.py
```diff
@@ -225,6 +225,71 @@ def point_dipole(omega, x0, n0, grid, *, c=None):
...
+        Sound pressure at positions given by *grid*.
+
+    Notes
+    -----
+    .. math::
+
+        G(\x-\x_0,\w) = \frac{1}{2\pi}
+            \frac{J_1(\wc R \sin(\Theta))}{\wc R \sin(\Theta)}
+            \frac{\e{-\i\wc|\x-\x_0|}}{|\x-\x_0|}
+    with :math:`\Theta` more conveniently defined by its cosine
```
mgeier: sfs/fd/source.py:docstring of sfs.fd.source.point_circular_piston:20: WARNING: Explicit markup ends without a blank line; unexpected unindent.

PR #157, opened 2019-11-22
add point source with line piston far-field directivity

fietew:


mgeier:
Same procedure as #156:

    $ python3 -m pycodestyle sfs/fd/source.py

Same Sphinx warning:

    sfs/fd/source.py:docstring of sfs.fd.source.point_line_piston:19: WARNING: Explicit markup ends without a blank line; unexpected unindent.

Issue #77, opened 2019-01-23
Find replacement for inner1d()


mgeier:
This deprecation warning appears on Travis-CI when using Python 3.7 (using Ubuntu Xenial):

```
sfs/mono/drivingfunction.py:8
  /home/travis/build/sfstoolbox/sfs-python/sfs/mono/drivingfunction.py:8: DeprecationWarning: numpy.core.umath_tests is an internal NumPy module and should not be imported. It will be removed in a future NumPy release.
    from numpy.core.umath_tests import inner1d  # element-wise inner product
```

I don't know what's the best replacement for `inner1d()`, suggestions and PRs are very welcome!

hagenw:
Here are some suggestions: https://github.com/numpy/numpy/issues/10815

spors:
One of the benefits of inner1d was parallel processing. This might not be the case for the replacements.

PR #163, opened 2020-03-11
Replace inner1d bei einsum

fs446:
We use
`from numpy.core.umath_tests import inner1d as _inner1d`
which is suboptimal, since its deprecated and requires non-standard library import.
See also https://github.com/sfstoolbox/sfs-python/issues/77

We could use numpy's `einsum` do to this job instead, thus I've changed (hopefully) all occurrences of `inner1d` to `einsum`.

Besides that in the files, some PEP8 modifications were done, these suggestions came from PyCharm.

Maybe numpy's `tensordot` is even more performant/suitable. I don't know currently how/if these functions support parallel computation, which `inner1d` seemed to do and was chosen for. I will check this.

In the meantime, using `einsum` in this PR is definitely an improvement for the code.

### DIFF
--- a/sfs/fd/wfs.py
+++ b/sfs/fd/wfs.py
@@ -32,7 +32,6 @@
 
 """
 import numpy as _np
-from numpy.core.umath_tests import inner1d as _inner1d
 from scipy.special import hankel2 as _hankel2
 
 from . import secondary_source_line as _secondary_source_line
@@ -91,7 +90,7 @@ def line_2d(omega, x0, n0, xs, *, c=None):
     k = _util.wavenumber(omega, c)
     ds = x0 - xs
     r = _np.linalg.norm(ds, axis=1)
-    d = -1j/2 * k * _inner1d(ds, n0) / r * _hankel2(1, k * r)
+    d = -1j / 2 * k * _np.einsum('ij,ij->i', ds, n0) / r * _hankel2(1, k * r)
     selection = _util.source_selection_line(n0, x0, xs)
     return d, selection, _secondary_source_line(omega, c)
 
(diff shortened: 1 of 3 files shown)


mgeier:
I like the PEP8 changes, except for the spaces around `**`.

> In the meantime, using `einsum` in this PR is definitely an improvement for the code.

"definitely" is quite a strong word. "improvement", too.

What if it's much slower?

mgeier:
I guess some instances can be replaced with the `@` operator (a.k.a. `__matmul__`). Have you checked the performance of that?

hagenw:
The test for Python 3.5 fails as the installed `sphinxcontrib-bibtex` package no longer supports Python 3.5:

```
Exception occurred:
899  File "/home/travis/virtualenv/python3.5.7/lib/python3.5/site-packages/sphinxcontrib/bibtex/roles.py", line 12, in <module>
900    import pybtex.database
901  File "/home/travis/virtualenv/python3.5.7/lib/python3.5/site-packages/pybtex/database/__init__.py", line 124
902    f"  entries={repr_entry},\n\n"
903    ^
904SyntaxError: invalid syntax
```
I also very much dislike Python 3.5 and would simply remove support for it.

hagenw:
The other test fails as `homebrew` seems not to be installing `pandoc`

hagenw:
I created a merge request where I run the tests directly on Github (and skip Python 3.5): https://github.com/sfstoolbox/sfs-python/pull/166
There the MacOS tests work as well. So it just seems to be a bug in how travis is configured.

hagenw:
I rebased the master, now the tests are passing.

fs446:
I rebased and made code work for recent 0.6.1 / 0.6.2. We will need a handling once inner1d() is discontinued, so it might be a good idea to keep this einsum() handling up to date. I would vote using this some time.

fs446:
we can close this, because #180 handled this in another way

PR #166, opened 2020-11-27
Replace travis with GitHub for running CI jobs

hagenw:
Since some time we can now run CI jobs directly with Github. I like that solution as you don't need to use an external service.
I would propose to switch our tests to Github as well.

In addition, this proposes to remove support for Python 3.5, which would bring the following benefits for writing the code:
* dictionaries behave in a more predictable way, as they are ordered since Python 3.6
* we can use `f'{variable} in strings'`

/cc @mgeier


hagenw:
BTW, the Github workflows can even be used to automate the package release process, e.g. for every tag [build wheel upload it to pypi.org](https://github.com/audeering/audiofile/blob/master/.github/workflows/publish.yml#L23-L34) and [create a release on Github from the CHANGELOG](https://github.com/audeering/audiofile/blob/master/.github/workflows/publish.yml#L48-L69).

.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
+name: Test
+
+on:
+  push:
+    branches: [ master ]
```
mgeier, line 5: Why limit this to `master`?

I think it's quite useful if I can run CI on my own branches before making a PR. This way, I can already do a few iterations with CI without annoying other collaborators and create the PR only when I already know that CI is passing.

I think if you limit the branch here, this is not possible, or is it?

.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
...
+  pull_request:
+    branches: [ master ]
```
mgeier, line 7: Why is this explicitly limited to `master`?
Won't all (or at least most) PRs be from `master`?
I don't really have experience with non-`master` PRs, but I don't see a reason why to prohibit them. Is there?

.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
...
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, macOS-latest ]
```
mgeier, line 15: Did you try Windows as well?

.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Prepare Ubuntu
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y pandoc ffmpeg
```
mgeier, line 27: I've recently seen the flag `--no-install-recommends`, which might reduce the number of unnecessarily installed packets?

I haven't tried it yet, though, so I don't know whether it actually makes any difference.

.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Prepare Ubuntu
+      run: |
+        sudo apt-get update
```
mgeier, line 26: Is this necessary?
I don't think it makes a big time difference, but I'm just curious why this is needed.

.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
...
+        python3 -m pip install -r doc/requirements.txt
+        # This is needed in example scripts:
+        python3 -m pip install pillow
+    - name: Test
+      run: python3 -m pytest
+    - name: Test documentation
+      run: |
+        python3 doc/examples/run_all.py
+        # This executes the example notebooks and runs the doctests:
+        python3 -m sphinx doc/ _build/ -b doctest
```
mgeier, line 47: I think it would be nice to have this as separate steps, wouldn't it?

mgeier:
Thanks, this is great!

I agree with removing Python 3.5 support, but you should reflect this in `setup.py`.

> BTW, the Github workflows can even be used to automate the package release process, e.g. for every tag [build wheel upload it to pypi.org](https://github.com/audeering/audiofile/blob/master/.github/workflows/publish.yml#L23-L34) and [create a release on Github from the CHANGELOG](https://github.com/audeering/audiofile/blob/master/.github/workflows/publish.yml#L48-L69).

I think uploading the package manually isn't a lot of work, but why not reduce the amount of work even more!

How would the PyPI authorization work?

Anyway, I think this should happen in a separate PR.

.github/workflows/test.yml
hagenw: I think it stems from me trying to limit CI jobs, as for example in private repos you can have only some hundereds each month.
But you are right, we can maybe completely remove the limit here.

.github/workflows/test.yml
hagenw: My hope would be that it never install recommended package,  but just recommends them. But I will check.

hagenw:
> Thanks, this is great!
> 
> I agree with removing Python 3.5 support, but you should reflect this in `setup.py`.

I removed it from `setup.py`
 
> > BTW, the Github workflows can even be used to automate the package release process, e.g. for every tag [build wheel upload it to pypi.org](https://github.com/audeering/audiofile/blob/master/.github/workflows/publish.yml#L23-L34) and [create a release on Github from the CHANGELOG](https://github.com/audeering/audiofile/blob/master/.github/workflows/publish.yml#L48-L69).
> 
> I think uploading the package manually isn't a lot of work, but why not reduce the amount of work even more!

If you have a lot of packages to maintain and different users that should be able to create releases it is helpful. If you only do a release once a month/year, it's indeed not really needed.

> How would the PyPI authorization work?
You can store [secrets](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets) in your personal or organisation account and use those to authenticate.

> Anyway, I think this should happen in a separate PR.

Yes.

.github/workflows/test.yml
hagenw: I removed the restriction.

.github/workflows/test.yml
hagenw: I removed all limitations and the CI jobs should now be executed for every push / pull request.

.github/workflows/test.yml
hagenw: I was wrong, there is [required, recommended, suggested](https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends) packages and only the suggested will not be installed.
So, it seems to be indeed a good idea to add `--no-install-recommends`, which I did now.

.github/workflows/test.yml
hagenw: No, this is not really needed. I removed it.

.github/workflows/test.yml
hagenw: I split it up into to sections, this way we also don't need the comment.

.github/workflows/test.yml
hagenw: Oh maybe it might be a problem, see https://github.com/sfstoolbox/sfs-python/pull/166/checks?check_run_id=1467403029
I think the test failed because it was not able to find the package as the index where to look for it (or for which version) was not up-to-date?

I have now reverted my commit to test again.

.github/workflows/test.yml
hagenw: Now it will run all CI jobs twice for pull requests, one time for the push event, one time for the pull request. This might be not necessary, or would you prefer it this way?

.github/workflows/test.yml
hagenw: I added tests for Windows as well now.

.github/workflows/test.yml
mgeier: If you know a way to avoid the duplication, that would be great. But I think both things individually are very useful.

.github/workflows/test.yml
mgeier: Good to know!

mgeier:
> > How would the PyPI authorization work?
>
> You can store [secrets](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets) in your personal or organisation account and use those to authenticate.

Yes, but who's secrets?
Would we need an organization account on PyPI?

I forgot, but we actually have one: https://pypi.org/user/spatialaudio/

Would we generate a secret from this account?

hagenw:
> > > How would the PyPI authorization work?
> > 
> > 
> > You can store [secrets](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets) in your personal or organisation account and use those to authenticate.
> 
> Yes, but who's secrets?
> Would we need an organization account on PyPI?
> 
> I forgot, but we actually have one: https://pypi.org/user/spatialaudio/
> 
> Would we generate a secret from this account?

Sounds reasonable. I'm not sure if you can see at https://pypi.org/project/sfs/0.6.0/ who did the actual publishing. Otherwise a private account would also work.

mgeier:
> I'm not sure if you can see at https://pypi.org/project/sfs/0.6.0/ who did the actual publishing.

I don't know. I didn't see any indication on the web page ...

> Otherwise a private account would also work.

What do you mean by private account?

hagenw:
Should have set personal PyPI account instead of private. For example at audeering we use the credentials of my PyPI account.

mgeier:
OK, thanks.

PR #167, opened 2020-12-01
Preparation of Release 0.6.0

spors:
Rendered: https://sfs-python--167.org.readthedocs.build/en/167/version-history.html

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -3,10 +3,8 @@ Version History
 
 
 Version 0.6.0 (2020-12-01):
-* Added spatially bandlimited line source
-* Extended documentation of various functions
-* Various bug fixes
-* Drop support for Python 3.5
+ * New function `sfs.fd.source.line_bandlimited()` computing the sound field of a spatially bandlimited line source
+ * Drop support for Python 3.5
 
 Version 0.5.0 (2019-03-18):
  * Switching to separate `sfs.plot2d` and `sfs.plot3d` for plotting functions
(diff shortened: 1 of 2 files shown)


NEWS.rst
```diff
@@ -1,6 +1,13 @@
 Version History
 ===============
 
+
+Version 0.6.0 (2020-12-01):
```
mgeier, line 5: You should check the generated HTML file, which is broken (https://sfs-python--167.org.readthedocs.build/en/167/version-history.html).

The bullet points have to be indented, like they are below.

NEWS.rst
```diff
@@ -1,6 +1,13 @@
...
+* Added spatially bandlimited line source
```
mgeier, line 6: It would probably be nice to add a link to the new function?

NEWS.rst
```diff
@@ -1,6 +1,13 @@
...
+* Extended documentation of various functions
```
mgeier, line 7: Sure, but do we really need to mention this, especially since its a totally generic statement carrying basically no information?

NEWS.rst
```diff
@@ -1,6 +1,13 @@
...
+* Various bug fixes
```
mgeier, line 8: Sure, I guess that's true for every single release, therefore it seems redundant to me.

spors: You are right, I will remove these generic statements.

mgeier (approved):

PR #168, opened 2021-01-16
introduce new wfs 2.5D point source time domain driving function

fs446:
As we introduced a new 2.5D WFS point source driving function in frequency domain with version 0.5.0 we should do this in time frequency as well for consistency, I'd propose.
Thus, for time domain same idea as for frequency domain: putting the new version into `point_25D()` (this is the 3D->2.5D Delft approach) and creating `point_25d_legacy()` that holds the function which was used up to <=0.6.0 (this is the 2D -> 2.5D approach by Rabenstein/Spors).

Furthermore, some minor corrections in the math docstrings for the other time domain WFS functions and some pep8 issues fixed.

I would suggest to create a 0.6.1 minor release to introduce the new feature.

PR #169, opened 2021-01-18
DOC: add compatibility for sphinxcontrib-bibtex>=2

hagenw:
the test in https://github.com/sfstoolbox/sfs-python/pull/168 are failing as `sphinxcontrib-bibtex` introduced a breaing change with version 2.0.0 as you have to add another config setting.

This fixes thie issue.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -46,6 +46,8 @@ extensions = [
     'nbsphinx',
 ]
 
+bibtex_bibfiles = ['references.bib']
+
 nbsphinx_execute_arguments = [
     "--InlineBackend.figure_formats={'svg', 'pdf'}",
     "--InlineBackend.rc={'figure.dpi': 96}",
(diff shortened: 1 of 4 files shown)


mgeier:
I think now the `references.bib` argument can be removed from `references.rst`:

```rst
.. bibliography:: references.bib
    :style: alpha
    :all:
```

Also, I think `:all:` isn't necessary anymore, right?

I'm not sure though, because I cannot check it locally because I'm getting some errors (which may be related to my broken installation).

hagenw:
Thanks, `references.bib` is indeed no longer needed. `:all:` includes all references, also not cited once. I removed that option as well.
I was able to execute the code locally and found also one wrong citation to Ahrens2012 that was not using the bibtex plugin, and corrected that as well.

mgeier (approved):
Thanks!

PR #179, opened 2024-09-16
Update to changes in matplotlib 3.9.0

MeggyCal:
This patch makes the testsuite pass even with the new matplotlib.


hagenw:
Thanks for providing this patch. In order to make the tests work again we need to cleanup their definitions:

* remove Python 3.6
* remove Python 3.7
* update read the docs configuration

Will try to handle this first in another pull request.

PR #191, opened 2025-09-02
Register more "clipped" color maps, use viridis_clip as default for level()

mgeier:
This also changes "alpha" to a slightly more extreme value. We could tune this for each colormap individually.

Alternatively, we could also manually select "over" and "under" colors for each colormap.

### DIFF
--- a/sfs/plot2d.py
+++ b/sfs/plot2d.py
@@ -8,27 +8,26 @@ from . import default as _default
 from . import util as _util
 
 
-def _register_cmap_clip(name, original_cmap, alpha):
+def _register_cmap_clip(name, original_name, alpha):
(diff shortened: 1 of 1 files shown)


mgeier:
The clipped colors can be seen e.g. here: https://sfs-python--191.org.readthedocs.build/en/191/sfs.td.nfchoa.html

For comparison, an older release: https://sfs-python.readthedocs.io/en/0.6.2/sfs.td.nfchoa.html

For these time-domain plots, most of the area has very low level and therefore has the "under" color.

I'm not sure yet whether that's a good thing or not.

mgeier:
I have added a few commits with a new way to calculate the under/over colors.

fs446:
Nice work, thank you very much!

PR #192, opened 2025-09-02
Add plot2d.level_contour()

mgeier:
This might come in handy for the plots in the 2.5D WFS examples from #184.


fs446 (approved):
Very nice new feature to draw level isobars, such as the 0 dB level.

putting stuff into plotting_plane () makes handling more convenient and potential re-work of this particular code is easier.

